PR #1, opened 2015-01-06
Add secretbox support

qwzybug:
Tried to keep to your code style as much as possible


jedisct1:
Oh, this is great, thanks a lot!

I'm very new to Swift, so if you think the style sucks and is not idiomatic Swift, your input would be more than appreciated.

Merged! Thanks again!

qwzybug:
Thanks! I'll definitely keep thinking about how to make crypto Swift-y. I don't mind the very straightforward style you currently use, though, it's very plain and clear, which is what you want from libraries like this.

PR #2, opened 2015-04-13
Add podspec for cocoapods

ziogaschr:
Thanks for the great work @jedisct1.

I have added cocoapods podspec in my fork. It will be nice if you can accept it.
When you start using tags, you can uncomment line 10 and remove line 9.


jedisct1:
Merged, thanks!

PR #3, opened 2015-04-15
Renamed podspec for cocoapods

ziogaschr:
Sorry for the confusion @jedisct1.

I have renamed the podspec from `swift-sodium` to `Sodium` as it will be easier for the user to use it as described in the documentation. 
Now dev can add `import Sodium`, otherwise he has to use `import swift_sodium`, which isn't so nice.

Also I wasn't sure about the license, so you might want to set the correct license like: `s.license     = { :type => "MIT", :file => 'LICENSE' }`

Also, you will have to push the podspec to the CocoaPods servers, as you are the owner.
For now users can use it like:
```
pod 'swift-sodium', git: 'https://github.com/jedisct1/swift-sodium.git', :branch => 'master'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,13 @@ script. Running this script on Xcode 6.1.1 (`6A2008a`) on the revision
 `cae09d458a4509614c6cdb32e93ab2f6263c18ce` of libsodium generates a file
 identical to the one present in this repository.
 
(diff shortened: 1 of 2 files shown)


jedisct1:
Thank you so much for your help, Chris!

PR #4, opened 2015-04-19
Added OSX Framework Target!!

ramki1979:
I have added OSX Framework build Target.


ramki1979:
In my local branch, I have added iOS and OSX app using the build frameworks.
But the app didn't have any thing than just 
```swift
import Sodium
let sodium = Sodium()!
```
If any one having some example code with UI that want to include, let me know, I will add the code and will make a pull request.

Or just let me know, I will merge the sample apps as it is and make a pull request.

ramki1979:
This is my First Pull request, I have not ask to merge the other two commits, I only pushed the other two commits to my repository. Why are they showing here as though I want to merge the example projects commits as well?

PR #9, opened 2015-09-15
Update for Carthage distribution

xNekOIx:
Please accept and add new tag (or move 0.1 to latest commit). :cactus:


Sodium/Sign.swift
```diff
@@ -38,7 +38,7 @@ public class Sign {
         return KeyPair(publicKey: PublicKey(data: pk!), secretKey: SecretKey(data: sk!))
     }
     
-    public func keyPair(seed seed: NSData) -> KeyPair? {
+    public func keyPair(#seed: NSData) -> KeyPair? {
```
jedisct1: This syntax doesn't exist any more since Swift 2.

xNekOIx: xcode 7 has not yet released. 
what is correct function signature will be in this case?
```public func keyPairWithSeed(seed: NSData) -> KeyPair?``` ?

jedisct1: `seed seed` works fine under Xcode 7, so I'd rather keep that since it also works on Xcode 6 minus a warning.

xNekOIx:
bump :cake:

jedisct1:
Done :)

xNekOIx:
:+1:

PR #10, opened 2015-09-16
Support for beforenm and afternm

guanix:
I am adding support for separate beforenm and afternm operates, for pre-generating the shared secret for slightly better performance and potential use of the shared secret in other contexts.


jedisct1:
This is great, thanks!

Do you mind adding tests for these?

guanix:
There are tests now.

jedisct1:
looks good! Thanks!

PR #15, opened 2015-10-15
Update README.md

stannie:
Attached signatures example should open/verify the signedMessage, not the original message


jedisct1:
Merged, thanks!

PR #19, opened 2015-12-15
added func overloads for anonymous seal and open

eliburke:
This PR adds seal() and open() funcs for crypto_box_seal and crypto_box_open, allowing one to anonymously send messages to a recipient given its public key. (docs here: http://doc.libsodium.org/public-key_cryptography/sealed_boxes.html )

I won't quibble too much but it did throw me for a loop that you used the seal verb without actually exposing libsodium's seal method. 😜

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,11 @@ Running these scripts on Xcode 7.0 (`7A220`) on the revision
 `437b20161ca0271756e5c2b6078a94dd63b51493` of libsodium generate files
 identical to the ones present in this repository.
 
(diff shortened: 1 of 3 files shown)


jedisct1:
Thanks! This is indeed a great addition.

Could you also possibly update the documentation and the tests?

eliburke:
I am not familiar with these words. "documen-what-shun"? "tehhssts"? 

Updated. And I patched in the guard syntax for consistency.

jedisct1:
Thanks!

PR #20, opened 2015-12-16
Adopts guard syntax for early exits

brunokoga:
Swift 2 introduced the guard keyword which can be used to early exits.
Adopting it makes the code a little bit cleaner (although less cheerful, without all the exclamation marks :)


jedisct1:
Indeed, guards are less cheerful, and I already miss all these exclamation marks :) But yep, this change is good and makes the code cleaner.

Thanks!

PR #21, opened 2016-01-21
Changes `InternalExtensions` to `NSData+Extensions`

brunokoga:
- The NSData and NSMutableData are now public (That's why I renamed the
  file)
- With this change, whenever we need to extend the lib, we can use the
  same methods as Sodium to access the data pointers instead needing to
  duplicate this code.


brunokoga:
Here is one example. To achieve this, I needed to duplicate the NSData/NSMutableData extensions to my own module.

```swift
extension SecretBox {
    public func seal(message: NSData, secretKey: Key, nonce: NSData) -> NSData? {
        if secretKey.length != KeyBytes {
            return nil
        }
        guard let authenticatedCipherText = NSMutableData(length: message.length + MacBytes) else {
            return nil
        }
        if crypto_secretbox_easy(authenticatedCipherText.mutableBytesPtr, message.bytesPtr, UInt64(message.length), nonce.bytesPtr, secretKey.bytesPtr) != 0 {
            return nil
        }
        return authenticatedCipherText
    }
}
```

jedisct1:
Looks good and useful!

Merged, thanks!

brunokoga:
man, you're fast. :)

jedisct1:
:)

PR #25, opened 2016-04-03
Fix warnings about missing includes in umbrella header

blochberger:
Fix warnings about missing includes in umbrella header

As these are Swift bindings I assume that the underlaying C headers don't need to be exposed to the final product that uses the Sodium framework.

The headers are needed to compile the actual framework which then exposes Swift modules, so this pull request moves all the C headers to the "Private" target.

Some headers are not even used directly by the Sodium framework, such as fe. `crypto_int32.h`. These headers are not included in the umbrella header and cause warnings during build:

> ⚠️ Umbrella header for module 'Sodium' does not include header 'crypto_int32.h'

This pull request moves headers, that are not needed directly by the Sodium framework to the "Project" target. This will stop the warnings to appear (you might need to clean the project first).

The following headers are causing warnings:
* `crypto_int32.h`
* `crypto_int64.h`
* `crypto_uint8.h`
* `crypto_uint16.h`
* `crypto_uint32.h`
* `crypto_uint64.h`
* and `crypto_sign_edwards25519sha512batch.h` which wasn't moved to the "Project" target and will still issue the warning.

According to [Project Editor Help: Setting the Visibility of a Header File][1]:

> Use these criteria to set the appropriate visibility of a header file:
>
> * Public: The interface is finalized and meant to be used by your product’s clients. A public header is included in the product as readable source code without restriction.
> * Private: The interface isn’t intended for your clients or it’s in early stages of development. A private header is included in the product, but it’s marked “private.” Thus the symbols are visible to all clients, but clients should understand that they're not supposed to use them.
> * Project: The interface is for use only by implementation files in the current project. A project header is not included in the target, except in object code. The symbols are not visible to clients at all, only to you.

Additionally all headers are sorted alphabetically in the project configuration (no source files changed) which makes it easier to spot specific headers.

[1]: https://developer.apple.com/library/ios/recipes/xcode_help-project_editor/Articles/Setting%20the%20Role%20of%20a%20Header%20File.html#//apple_ref/doc/uid/TP40010155-CH9-SW1


jedisct1:
This is great!

Thanks you so much!

Merged.

PR #27, opened 2016-05-03
Fix examples

blochberger:
Unfortunately jedisct1/swift-sodium#26 didn't fix all examples, so that's what this is for.

The examples could also be implemented in unit tests to ensure that they're up to date and valid. They should go into a separate file, as `toData()` is defined as a `String` extension in the existing unit tests.


jedisct1:
Thanks!

PR #30, opened 2016-07-13
Fix examples

blochberger:
Sorry, I actually broke the examples in a7942cbbf3e4046296951dd6ae8bb150c2ab73db.

I missed that some methods have ambiguous return types, for which the type can not be inferred automatically. Fixed that by specifying the actual expected return type.

Additionally I added initialization calls to each example code snippet to ease copy-paste from examples and make them more uniform.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ Secret-key authenticated encryption
 ===================================
 
 ```swift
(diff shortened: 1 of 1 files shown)


jedisct1:
Merged, thanks!

PR #32, opened 2016-07-28
Add stream crypto chacha20 and salsa20

alex-chan:
Just use it like so:
```
        let key = String(count: 32, repeatedValue: Character("k")).dataUsingEncoding(NSUTF8StringEncoding)
        let iv =  String(count: 32, repeatedValue: Character("i")).dataUsingEncoding(NSUTF8StringEncoding)
        
        sodium.stream.initStream(key: key!, iv: iv!)
        
        let encrypted = sodium.stream.update(message)
        print("Encrypted with chacha20:\(encrypted)")
        
        let message2 = "hohohahi rezhewudi".dataUsingEncoding(NSUTF8StringEncoding)
        let encrypted2 = sodium.stream.update(message2!)
        print(encrypted2)
        
        let sodium2 = Sodium()!
        sodium2.stream.initStream(key: key!, iv: iv!)
        print("Decrypted chacha20:\(sodium2.stream.update(encrypted!)!.toString())")
        print("Decrypted chacha20:\(sodium2.stream.update(encrypted2!)!.toString())")
        
```


jedisct1:
Hi Alex,

And thanks for proposing this!

Bindings for `crypto_stream` were intentionally not present, because it is very easy to misuse, and `crypto_box` or `crypto_aead` is usually what people actually need.

As an illustration, your example above should mention that the ciphertext is not authenticated; anyone can alter the data, and it will not be detected. 

Huge data blobs should rather be split into blocks, each with an authentication tag that `crypto_box` and `crypto_aead` provides.

Low-level APIs to access the raw stream ciphers are mainly only useful if you want to build your own crypto. Having access to `crypto_stream` can be useful in this context, and might be a worthy addition to the bindings. But a higher-level abstraction that actually provides no security is dangerous and should probably not be added.

alex-chan:
Hi jedisct1: 
I didn't get it. you mean `crypto_box` did the same thing like function `crypto_stream_chacha20_xor_ic` ?

jedisct1:
`crypto_stream_xor` = not secure (you usually also need to compute a MAC).
`crypto_box` and `crypto_aead` = secure.

For a stream, it's better to split the data into small chunks, individually encrypted and authenticated using `crypto_box` or `crypto_aead`. 

See https://www.imperialviolet.org/2014/06/27/streamingencryption.html for details, or how related projects are doing it:

- https://github.com/bitbeans/StreamCryptor
- https://github.com/dominictarr/pull-box-stream
- https://github.com/jedisct1/blobcrypt

alex-chan:
Oh!  thank you for your patient answer, please close this pull request.

PR #34, opened 2016-08-16
Fix Swift 3 compiler errors

alexzautke:
Hi there!

In order to successfully compile swift-sodium using Xcode 8 (8S193k) a few changes were necessary:

1. Fix some renamed method names
2. Fix argument labels. Now, the first parameter of an method must have an argument label.
E.g. seal(message, secretKey: secretKey) must be written as seal(message: 
message, secretKey: secretKey)
3. The data argument of the NSMutableData initialiser uses the new "Data" type => Fix cast errors
4. Replace dispatch_once_t with a static var (which uses dispatch_once_t in the background)

All test are still passing.
Maybe it's worth considering to open a second branch besides the Master, which contains a Swift 3 compatible version of swift-sodium?

### DIFF
--- a/Sodium/Box.swift
+++ b/Sodium/Box.swift
@@ -72,7 +72,7 @@ public class Box {
     }
     
     public func seal(message: NSData, recipientPublicKey: PublicKey, senderSecretKey: SecretKey) -> NSData? {
-        guard let (authenticatedCipherText, nonce): (NSData, Nonce) = seal(message, recipientPublicKey: recipientPublicKey, senderSecretKey: senderSecretKey) else {
+        guard let (authenticatedCipherText, nonce): (NSData, Nonce) = seal(message: message, recipientPublicKey: recipientPublicKey, senderSecretKey: senderSecretKey) else {
             return nil
         }
         let nonceAndAuthenticatedCipherText = NSMutableData(data: nonce as Data)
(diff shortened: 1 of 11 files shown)


jedisct1:
Thanks a lot, Alexander!

Maintaining a separate branch until Xcode 8 is out of beta is indeed probably the way to go, as having the same code compile on both doesn't seem to be an option.

blochberger:
The examples should be transformed to Swift 3 as well.

jedisct1:
Merged to a `swift-3` branch until Xcode 8 final is released.

Thanks a lot, Alexander!

alexzautke:
There were still some errors in Xcode version 8.0 beta 6 (8S201h).
The three latest commits fix these errors.

The most notable change is, that I changed the computed property ''bytesPtr'' into a generic function. Through this new function it's possibly to get a Unsafe[Mutable]Pointer for any data type from an Unsafe[Mutable]RawPointer. As of the newest version of Xcode 8.0 NSData.bytes and the corresponding method of NSMutableData return a raw pointer. This raw pointer must be cast to a typed pointer before it can be used by libsodium.

The function can be called in the following way:
let pointer : UnsafeMutablePointer<Type> = output.mutableBytesPtr()

or let the compiler infer the correct type by passing the function as a parameter.

jedisct1:
Merged, thanks!

Issue #40, opened 2016-10-14
Swift 3 Support 


pgbear:
I created a test project with sodium library added using cocoapods. I was able to install version 0.2.0. But when I tried to build XCode 8 throws swift versioning errors. I can manually fix those errors to build the project but is it possible to fix them ahead 
[Test.zip](https://github.com/jedisct1/swift-sodium/files/528549/Test.zip)
?

blochberger:
Do the suggestions made in #38 help?

pgbear:
Great! I had to mention git branch as well configure to not use the legacy swift version.

`pod 'Sodium', :git => 'https://github.com/jedisct1/swift-sodium.git'`
```
post_install do |installer|
    installer.pods_project.targets.each do |target|
        target.build_configurations.each do |config|
            config.build_settings['SWIFT_VERSION'] = '3.0'
        end
    end
end
```

Issue #38, opened 2016-10-01
Sodium Pod Swift 3


gpalomar:
Hi!
I'm downloading the Sodium library using CocoaPods.
`source 'https://github.com/CocoaPods/Specs.git'`

`platform :ios, '8.0'`
`inhibit_all_warnings!`

`use_frameworks!`

`target 'my_app' do`
`    pod 'Fabric'`
`    pod 'Crashlytics'`
`    pod 'Sodium'`
`end`

All installed correctly but when I launch the `.xcworkspace`Xcode8 shows me this alert:
<img width="422" alt="captura de pantalla 2016-10-01 a les 13 50 49" src="https://cloud.githubusercontent.com/assets/3463553/19013889/1b6dbc58-87de-11e6-9135-2c5735fff63b.png">

I don't understand why this happens because in repository readme says **Swift 3 is now the default**.

What can I do?

Thanks!

ChaosCoder:
It's because there was not a new release version of this library, so the pod still refers to an old commit (where swift-3 was not yet merged). 

Until a new version is released you could tell your Podfile to use the current version on the master branch by just defining:
```
pod 'Sodium', :git => 'https://github.com/jedisct1/swift-sodium.git'
```

gpalomar:
Thanks for the reply @ChaosCoder 
This was what I thought, but using this wait to get the framework, I'm still getting the same problem. Xcode tells me that Sodium shoud be unlocked to convert it to Swift 3.

After days of silence in this post, I'm sure that it's my fault, but I don't know how to resolve it :worried:

ChaosCoder:
Did you set the library to use swift 3 as compiler setting?

Both top answers can be used, the top one is permanent, the second one has to be done after each `pod install`: http://stackoverflow.com/questions/38446097/xcode-8-beta-3-use-legacy-swift-issue

gpalomar:
@ChaosCoder you are the MVP! :smiley:

Using [this](http://stackoverflow.com/a/38466703), Sodium works perfectly in Xcode8.
Thanks a lot!!

Issue #37, opened 2016-09-15
Support to Xcode 8 and Swift 3


casevictor:
Hello folks, 
I know that @jedisct1 is actually working officially at the _swift-3_ branch, and it will be release soon.
But for anyone else that need the support for "yesterday" I took the freedom to create a fork with full support to Xcode 8 and Swift3. I did some tests and tricks today and everything seems fine..

Check the repo
https://github.com/casevictor/swift-sodium/tree/swift-3

@jedisct1 thanks for the amazing job!

jedisct1:
The master branch is now Swift-3 only.

Thanks!

gpalomar:
Hi @jedisct1 !
I'm trying to install swift-sodium using cocoapods and all the code it installs is not in swift 3. Can you explain me what I'm doing wrong and how can I fix it?

If it helps, my Podfile looks like this:
`platform :ios, '8.0'
inhibit_all_warnings!
use_frameworks!
 
target 'myApp' do
    pod 'Fabric'
    pod 'Crashlytics'
     pod 'Sodium'
end`

Thanks Frank!

PR #42, opened 2016-10-19
Disable Swift legacy version due to change to Swift 3

blochberger:
Although the syntax is Swift 3 already the Xcode project was left out while switching to the new Swift version. This lead to Xcode prompting the user to upgrade the syntax to Swift 3 upon opening the project. It also lead to the project not building (the iOS targets) due to a dependency analysis error:

> “Use Legacy Swift Language Version” (SWIFT_VERSION) is required to be configured correctly for targets which use Swift. Use the [Edit > Convert > To Current Swift Syntax…] menu to choose a Swift version or use the Build Settings editor to configure the build setting directly.

This patch fixes this behaviour.

This is related to #37, #38, and #40.


mmackowiak:
+1

jedisct1:
Sure!

PR #43, opened 2016-10-19
Fix warnings for unused results

blochberger:
This removes three warnings from the build, and it's probably not bad to check the result there to see if the operation succeeded.

The warnings were:

> Result of call to update(input:) is unused.


jedisct1:
Merged, thanks!

PR #44, opened 2016-10-20
Apply Xcode 8 recommendations

blochberger:
Since Swift 3 is now the default and Xcode 8 is required for that, we might as well increase the Xcode version. Now that nearly all recommendations for the newer Xcode version are applied it will avoid Xcode performing update checks upon startup.

The only recommendation that I did **not** apply was to disable code signing for the dynamic libraries during build phase. I'm not really sure why Xcode says that it is not recommended and it might need some further investigation/discussion.

The applied recommendations are some additional warnings and whole module optimization being enabled.


mmackowiak:
Looks good. :)

jedisct1:
Makes sense!

Merged, thanks!

Issue #24, opened 2016-03-27
Create Documentation


blochberger:
I know there is [documentation for libsodium](https://download.libsodium.org/doc/) but it would be nice to have some documentation for the Swift classes that integrates nicely with Xcode, see [Swift Documentation](http://nshipster.com/swift-documentation/).

jedisct1:
Thanks, @jnross !

jnross:
I confirmed that these docs make it through the Carthage build process and appear in my project that consumes the framework :)

jedisct1:
Excellent :)

PR #48, opened 2016-12-12
Update all interfaces to use Swift 3's new Data type

nathankot:
Swift is moving towards `Data` as a replacement for `NSMutableData` and
`NSData`, this PR simply adds support for the `Data` type in the
utils. The primary difference between `Data` and the legacy types is
that memory isn't guaranteed to be contiguous, therefore
`withUnsafeBytes` should be used in order to get a pointer into the
`Data` as opposed to using `data.bytes`.

This also makes a *potentially breaking* change in that `zero()` no
longer resets the data length to `0`, happy to revert this but to me
'zero-ing' memory means reseting the bits to `0`, not the actual length
of the data - this makes for some nicer tests as well since we can check
that the underlying bits are actually zero.


nathankot:
Closing, just realized I should have forked from the `swift-3` branch.

nathankot:
Seems like the `swift-3` branch has been merged and then diverged https://github.com/jedisct1/swift-sodium/compare/master...swift-3 with `swift-3` using the new `Data` type, is this project planning to the new `Data` type for its interfaces?

blochberger:
I think the proposed changes should go in. I was already planning to do a PR for that myself.
As mentioned in https://github.com/jedisct1/swift-sodium/issues/37#issuecomment-248006523 the `master` branch is now Swift 3 only and I assume that the `swift-3` branch is deprecated.

It's good that the tests with `NSData` are kept to make sure that current implementations are still compatible, but I would add tests using the `Data` class as well, just to make sure – especially if their memory layout is different.

I like actually being able to test/validate that the memory was zeroed.

jedisct1:
These changes are great, thanks!

Would you mind also updating the documentation?

Thanks again for your contribution!

nathankot:
⚡️ `README.md` updated, although the other samples and example projects have all diverged from the swift 3 changes and can do with a proper cleanup. Don't think it's within the scope of this PR but I'm happy to look into getting full affordance of `Data` and clean up the examples if there hasn't been progress on this :)

blochberger:
I think the documentation update is incomplete,

```swift
let message = "My Test Message".dataUsingEncoding(NSUTF8StringEncoding)!
```

should look like:

```swift
let message = "My Test Message".data(using: .utf8)!
```

Maybe we should transform the examples to unit tests to quickly see if they compile (in another PR).

nathankot:
> I think the documentation update is incomplete,

Yup, just to re-iterate:

> although the other samples and example projects have all diverged from the swift 3 changes and can do with a proper cleanup

Since most of the examples/docs have already diverged, I think it makes sense to do this in a separate PR :)

nathankot:
@jnross thanks for updating the `README.md` examples! I've rebased this branch off that and updated *all* interfaces to use the new `Data` type, enjoy :)

cc @jedisct1 @blochberger

jnross:
@nathankot: This PR looks great! I look forward to it being merged.  I was working on a similar set of changes which turned out to be a great exercise and taught me a lot about the codebase.  But your changes are far more complete and than where I was, so I'll just work from the head of your PR for now.

The only potential change I'm thinking about now is whether first argument names should be removed from some methods like they were back in Swift 2.  The various`seal(message:String,...` methods in particular feel like they don't need the first argument label `message:`.  What do you & @jedisct1 think?

blochberger:
I personally think that labels for arguments improve readability of code. If the first argument is removed an Objective-C / Swift 2 naming style would be more helpful, f.e. `sodium.box.sealMessage(msg, …)` vs. `sodium.box.seal(msg, …)`. To avoid confusion with the documentation on [libsodium.org](https://libsodium.org) due to different function names used in the first case, naming the first argument would be my preferred option, unless there is separate documentation (requested in #24).

jnross:
@blochberger: Alright, sounds like this warrants another conversation that is out of scope for this PR.  My apologies for the digression.  I hope to see this PR land soon and then we can discuss first argument names in another PR.

nathankot:
@jnross @blochberger 

> I personally think that labels for arguments improve readability of code.

Personally I echo this sentiment, and agree it's a discussion for another PR, thanks for getting this merged everyone!!

PR #53, opened 2017-01-16
Make Sodium() initializer nonfailable.

jnross:
The Sodium initializer never actually returns nil, so I've marked it as unfailable.  This will remove the need for a forced unwrap or `if let` or `guard let` in code that uses this framework.

The other alternative would be to actually return nil if the call to `sodium_init()` fails.  I've prepared a commit that does that here: 55782c983d2b59a85e6e14e58cc2914ef76db63e.  There are other approaches to this since it is safe to call `sodium_init()` multiple times.

However, I prefer switching the existing implementation to a nonfailable initializer, allowing it to fatal error if `sodium_init()` fails.  Failure is very unlikely and difficult to recover from, and having the nonfailable initializer allows for tidier use of the Sodium object.

### DIFF
--- a/Examples/OSX/AppDelegate.swift
+++ b/Examples/OSX/AppDelegate.swift
@@ -18,7 +18,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     func applicationDidFinishLaunching(aNotification: NSNotification) {
         // Insert code here to initialize your application
 
-        let sodium = Sodium()!
+        let sodium = Sodium()
         let aliceKeyPair = sodium.box.keyPair()!
         let bobKeyPair = sodium.box.keyPair()!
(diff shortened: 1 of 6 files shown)


blochberger:
Please see discussion for commit d3432d63a095d301a30073fb69fcfbdb75ec7128.

jnross:
@blochberger: Thanks for pointing me to that discussion.  I tend to agree with some of your remarks in that thread. What it comes down to is that having this initializer be failable provides no value as currently implemented (at least none that I can see).  If the initializer is failable, then it should return nil in the case that initialization causes failure - this will allow the host program to handle the failure at it sees fit rather than crashing with `abort()`.  I'll close this PR and open another with the other solution I suggested.

PR #54, opened 2017-01-16
Make Sodium init properly failable.

jnross:
There are a couple of ideas captured here:

First, if the initializer is failable, it should actually return `nil` when failure occurs, not crash the host application.  Returning nil gives the host application the opportunity to handle the error as it chooses.  It may be that many will choose to trigger a fatal error:
```
guard let sodium = Sodium() else {
    abort()
}
```
But the choice must be available.

Second, this commit prepares for the case in which a failed call to `sodium_init()` may recover in subsequent calls.  From the [libSodium docs](https://download.libsodium.org/doc/usage/):
> `sodium_init()` initializes the library and should be called before any other function provided by Sodium. The function can be called more than once, and can be called simultaneously from multiple threads since version 1.0.11.

> `sodium_init()` returns 0 on success, -1 on failure, and 1 if the library had already been initialized.

### DIFF
--- a/Sodium/Sodium.swift
+++ b/Sodium/Sodium.swift
@@ -19,12 +19,8 @@ public class Sodium {
     public var utils = Utils()
 
     public init?() {
-        struct Once {
-            static var once : () = {
-                if sodium_init() == -1 {
-                    abort()
-                }
-            }()
+        if sodium_init() == -1 {
+            return nil
         }
     }
 }


blochberger:
Since there already is a singleton kept behind `sodium_init()` I think that removing the superfluous singleton is a good idea. Since `sodium_init()` can be called simultaneously I do not see a need for protecting the call with the `once` token.

It might affect some corner cases, e.g. binding `libsodium.a` multiple times dynamically, where each C-side singleton has it's own address space, but I'm not sure if that is a real concern. You probably have other problems then anyway.

Using `let sodium = Sodium()!` would still lead to a crash, although I prefer the `guard`ed `abort()`, personally due to better readability and more accurate error messages.

blochberger:
Just for future reference: see discussions in d3432d63a095d301a30073fb69fcfbdb75ec7128 and #53.

zetasq:
I suggest not using failable initializer, because the client normally cannot handle the situation when **sodium_init()** return -1. We could just call **fatalError()** and print the message. A non-optional initializer seems more elegant.

By the way, I don't think we need to call sodium_init() in every init. I think the following code is better:

```
public class Sodium {
    private static let once = {
        if sodium_init() == -1 {
           fatalError("Failed to init libSodium")
        }
    }()

    public init() {
        _ = Sodium.once
    }
}
```

joejune:
@zetasq:  I think a non-failable initializer which `fatalError`s would be fine too.  That approach vs the one outlined in this PR are basically two sides of same coin:  Should the library trigger the fatal error, or should that be left to the integrator.

My primary point is that the current implementation brings confusion:  it is marked as failable, but never returns nil and may `abort`.

jnross:
@zetasq:  I think a non-failable initializer which `fatalError`s would be fine too.  That approach vs the one outlined in this PR are basically two sides of same coin:  Should the library trigger the fatal error, or should that be left to the integrator.

My primary point is that the current implementation brings confusion:  it is marked as failable, but never returns nil and may `abort`.  In other words, I support the changes you propose in #106 as an alternative to the changes proposed here.

zetasq:
@jnross Ah...I just closed that pull request yesterday. Should I recreate the pull request?

joejune:
@zetasq:  It's up to you.  This particular PR has been open for almost nine months, so I'm not really expecting a change at this point.

jnross:
@zetasq: It's up to you. This particular PR has been open for almost nine months, so I'm not really expecting a change at this point.

jedisct1:
I'm totally fine with such a change! I was just waiting for opinions to converge :)

blochberger:
@jedisct1, are there any non-critical issues you expect to occur during initialization? I still vote for a non-optional and if there is a change in the future that is non-critical it might force users/callers to think carefully how to handle the problem.

jnross:
Back in January, I think myself and @blochberger were pretty well aligned that making this properly failable was a good choice.  The advantage of this is that it allows the integrator to choose how to handle the failure.  It is likely that in many cases the failure would be effectively fatal, but it gives a chance to message the user about what went wrong or what can be done to resolve the situation.

@zetasq seems to prefer a non-failable initializer which calls `abort()` within it.  This means that the integrator can't control the failure.  But since failure is very unlikely, this approach has the great advantage that `Sodium()` can be initialized many times without having to unwrap the optional each time.

I'm ok with both approaches.  Either is much better than the current implementation.

Updated:  Seeing @blochberger's recent comment, it sounds like most are in favor of a non-failable initializer.  Therefore I suggest we reopen and merge #106, and I'll be happy to close this one.

zetasq:
@jnross I just re-create my pull request #108 (I can't re-open it because I deleted my fork after closing the previous pull request).

blochberger:
This can be closed since #108 was merged.

jedisct1:
Closed!

jnross:
🎉

Issue #58, opened 2017-02-08
Diffie-Hellman Key Generation Support


weyhan:
Hi,

Is there any plans to support Diffie-Hellman key generation for key exchange? Any timeframe if there's plans?

Thanks.

jedisct1:
Closing because the plan is detailed in #60 :)

PR #60, opened 2017-02-21
Add key exchange

ChaosCoder:
This adds the key exchange functionality of libsodium to swift-sodium.

As I wanted to keep the structure of lib sodium, this PR introduces a new class `KeyExchange` with the following method:
```
func diffieHellman(secretKey: Box.SecretKey, publicKey: Box.PublicKey, otherPublicKey: Box.PublicKey) -> Data?
```
which takes the parameters:

`secretKey`: The secret key to use in diffie hellman key exchange
`publicKey`: The public key used in conjunction with the other public key to hash the computed secret
`otherPublicKey`: The other user's public key to use in diffie hellman key exchange

I added a test for this method and updated the README to contain a code snippet.
This also solves #58.

For more information on the key exchange via Diffie-Hellman, have a look at [Libsodium reference](https://download.libsodium.org/doc/advanced/scalar_multiplication.html).

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,24 @@ if sodium.pwHash.strVerify(hash: hashedStr, passwd: password) {
 }
 ```
 
+Key exchange (Diffie Hellman)
+=============================
+
+```swift
+let sodium = Sodium()!
+let aliceKeyPair = sodium.box.keyPair()!
+let bobKeyPair = sodium.box.keyPair()!
+
+let sharedSecretForAlice = sodium.keyExchange.diffieHellman(secretKey: aliceKeyPair.secretKey,
+                                                            publicKey: aliceKeyPair.publicKey,
+                                                       otherPublicKey: bobKeyPair.publicKey)
+let sharedSecretForBob = sodium.keyExchange.diffieHellman(secretKey: bobKeyPair.secretKey,
+                                                          publicKey: bobKeyPair.publicKey,
+                                                     otherPublicKey: aliceKeyPair.publicKey)
+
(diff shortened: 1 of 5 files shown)


jedisct1:
Hi,

And thanks for that pull request!

A similar function was introduced a long time ago in the PHP extension. As far as I know, nobody used it. Ever.

A couple things may explain it:
- It's only available in this extension. The same key exchange cannot be done in another language without reimplementing the exact same construction.
- It's not sufficient to build an actual protocol.

On the other hand, being able to perform a key exchange is a critical feature to have.

In order to solve this, my plan was to implement (directly in libsodium) a generic API that can be used to perform complete key exchanges.

I just added such an API to libhydrogen: https://github.com/jedisct1/libhydrogen/wiki/Key-exchange

The intent is to add something very similar to libsodium, so that all bindings can leverage the same key exchange mechanism.

ChaosCoder:
Okay, feel free to do whatever with this PR. I actually need a DH key exchange for my current project and as I'm only using this library currently on all ends, I will just use my fork from now on.

jedisct1:
I understand that key exchange is really something we need shortly, including in the Swift wrapper.

Your PR sorts the public keys. This is fine, but out of curiosity, why did you feel that it was necessary?

Are these use cases where an application doesn't know if it initiated a connection or if it is acting as a server? How do you protect against nonce reuse in such a scenario?

jedisct1:
Version 1.0.12 includes a DH API: https://download.libsodium.org/libsodium/content/key_exchange/

ChaosCoder:
Nice, will you write a wrapper for it?

jedisct1:
If you want to help and contribute it, that would be awesome :)

PR #63, opened 2017-03-27
Sodium.h OS X target membership public=>private

caravaneer404:
This attribute make my CI fails during carthage bootstrap.


jedisct1:
Thanks!

PR #73, opened 2017-04-28
Use Swift module rather than umbrella header for bindings

blochberger:
Details are explained in a [StackOverflow answer][1], which unfortunately is not accepted at the time of this commit.

This also separates all of the original `sodium` library files from the Swift bindings.

Now we need to `import libsodium` in every Swift file, but there is no need to manually manage the headers of the original library anymore.

[1]: http://stackoverflow.com/a/37072619/5082444


jedisct1:
Awesome, thanks!

Issue #75, opened 2017-05-05
Bug: Current Podspec for Sodium/libsodium is broken


ricwein:
Since the commit 865e819616b36100318378638bfe106fbaa0e48c moves all libsodium files into a subdirectory, the Podspec should adapt the changes. Using the current Podspec will skip all files inside Sodium/libsodium directory, resulting in missing `libsodium-ios.a`/`libsodium-osx.a` and more

Instead of

```ruby
s.ios.vendored_library    = 'Sodium/libsodium-ios.a'
s.osx.vendored_library    = 'Sodium/libsodium-osx.a'

s.source_files = 'Sodium/*.{swift,h}'
```

the lines should probably be

```ruby
s.ios.vendored_library    = 'Sodium/libsodium/libsodium-ios.a'
s.osx.vendored_library    = 'Sodium/libsodium/libsodium-osx.a'

s.source_files = 'Sodium/*.{swift,h}', 'Sodium/libsodium/*.{swift,h}'
```

But even then, building with xcode for iOS fails, being unable to import libsodium.

Greetings

blochberger:
Thanks for reporting, can you provide a pull request with the suggested changes?

What target fails to build: Your own application that links against the `Sodium` framework or the Sodium project?

ricwein:
Added a PR #76 

Building the Sodium Target, which was added through Pods fails to build. I just created a new test-project, but still... `import libsodium` causes an "No such module 'libsodium' error.

![screen shot 2017-05-05 at 17 17 26](https://cloud.githubusercontent.com/assets/870354/25751869/c31467b8-31b6-11e7-809a-b853858016c5.png)

At least the libsodium-ios.a exists again with the changed podspec.

ricwein:
With PR #78 getting merge, the build issues should be gone. Thanks for your help!

psalami:
If anyone is still having this issue even after updating to the latest version:
I needed to upgrade from Cocoapods 1.1.1 to 1.2.1 in order for this to work. Otherwise still seeing
``No such module 'libsodium'``

PR #76, opened 2017-05-05
fixes podspec source paths

ricwein:
see issue #75 

This pull-request should fix missing files by using Pods, since the libsodium core was moved from `Sodium` to `Sodium/libsodium`.


jedisct1:
Merged, thanks!

PR #78, opened 2017-05-06
fixes build-issues using Cocoa Pods

ricwein:
Sadly the proposed solution from #77 did not work for Cocoa-Pods, but the Idea was the right one.
The Podspec file just had to adapt the fixed `SWIFT_INCLUDE_PATHS`.

This should now solve all build issues with CocoaPods. Original issue see #75.


jedisct1:
Thanks!

PR #81, opened 2017-05-15
add `Box.open()` overload that takes a user-specified nonce

opfeffer:
It's been a breeze to work with swift-sodium. Thanks so much for your work.

I am working on a BLE encryption for which the peer device requires a very specific nonce. This PR adds the ability to swift-sodium's `Box` for a user to provide his own nonce instead of relying on `self.nonce()` to generate a random encryption nonce.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,9 @@ let messageVerifiedAndDecryptedByBob =
 
 `seal()` automatically generates a nonce and prepends it to the
 ciphertext. `open()` extracts the nonce and decrypts the ciphertext.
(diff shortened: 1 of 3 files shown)


jedisct1:
Hi!

And thanks for using this library!

In order for this to be merged, would it be possible for you to update the test suite and documentation to include the optional explicit nonce?

Thanks!

opfeffer:
Updated tests and documentation. Let me know if thats enough detail you were looking for.

jedisct1:
Merged, thanks!

PR #83, opened 2017-05-17
Fixes Xcode warnings re: differing file paths in cocoapods-generated umbrella header

opfeffer:
Xcode 8.3.2 and Cocoapods 1.2.1 throw the following warning when installing `Sodium` via cocapods:

warning

     In file included from <module-includes>:1:
     ~/SodiumTest/Pods/Target Support Files/Sodium/Sodium-umbrella.h:69:9: Non-portable path to file '"Sodium.h"'; specified path differs in case from file name on disk

From what I can tell, the problem stems from the fact that there's an capitalized `Sodium.h` which serves as the framework header for the actual swift framework, as well as a lowercased `sodium.h` which appears to be the module header of the `libsodium` module.

To resolve the warning, I marked `sodium.h` as private in the podspec to prevent cocoapods from including it in its auto-generated umbrella header.

Resolving this warning is important for projects which use build setting "GCC_TREAT_WARNINGS_AS_ERRORS".

Let me know what you think.  Thanks


jedisct1:
But why remove `sodium_lib.h` from the project?

blochberger:
I have renamed `sodium_lib.h` to `sodium.h` in 865e819616b36100318378638bfe106fbaa0e48c, which causes the addressed problem.

The removals in `Sodium.xcodeproj/project.pbxproj` were probably done by Xcode, due to a reference to a missing file – as I forgot to update the references there. This means that the `sodium.h` file is not listed in the "Supporting Files" group in the Xcode project.

I could revert the rename, which would also solve the issue this pull request solves originally. Although it might be wise to accept this pull request, fix the reference in another one and restructure the `Sodium/libsodium` directory in a way, that the includes can be copied from [jedisct1/libsodium](https://github.com/jedisct1/libsodium) directly.

jedisct1:
The original `sodium_lib.h` file is not the same as the one from libsodium, since it doesn't include files from system-defined paths. Is there a way to address that without breaking Carthage/Cocoapods?

If not, I would revert the rename. And still, having two files named `Sodium.h` is not great, so maybe reverting the rename may not be bad in either case.

blochberger:
I reverted the rename in #84.

Do the headers in `Sodium/libsodium` need to be listed in the umbrella headers? Since 6845200f10954a1514c162a70e480273886e8318, `libsodium` should be a Swift module in itself and therefore bridging from Objective-C should not be required anymore.

opfeffer:
Yea, I had only removed references to missing files on disk from the xcode project. since it was working without, I didn't question it much. Figured I'd do some house-cleaning haha. Apologies.

Max brings up a good point, should the `libsodium` headers be public in the first place. That essentially exposes all the underlying methods to the parent project. Ideally, you'd just want to expose the swift-friendly SwiftSodium methods, right?

In which case, it might make sense to mark all headers in Sodium/libsodium/*.h private? Happy to adjust my PR if thats the route you'd like to go.

jedisct1:
Keeping the `libsodium` headers private totally makes sense.

opfeffer:
I updated this PR, marking all headers in `Sodium/libsodium` private in the *.podspec

I also renamed sodium.h back to sodium_lib.h and added it back to the project file (essentially what #84 does).

Take a look?

jedisct1:
Looks good!

Thanks a lot, Oliver!

Merged!

blochberger:
Looks good.

jedisct1:
Thanks to both of you!

PR #84, opened 2017-05-17
Avoid having two files with the same name (case-insensitive)

blochberger:
This resulted in portability warnings for Cocoapods-generated umbrella headers (see #83).

This partially reverts commit 865e819616b36100318378638bfe106fbaa0e48c.


blochberger:
Solved by #83 already.

PR #89, opened 2017-06-15
Replace libsodium-ios.a (+armv7s)

Vladlex:
> The Sodium library itself doesn't have to be installed on the system: the repository already includes a precompiled library for armv7, armv7s, arm64, as well as for the iOS simulator.

Current libsodium-ios.a does not have armv7s arch slice (I've checked it by lipo -info libsodium-ios.a).

Here is rebuilt binary with armv7s slice, built from original sources using original script (ios.sh).


jedisct1:
Sorry, I'm not going to merge pull requests containing binary code unless they come from a government agency along with a very big check.

More seriously, this has been updated, if only to sync the precompiled libraries with the current stable version, and use the latest stable Xcode version as well.

Thanks!

Vladlex:
Thanks for the quick fix!

Issue #93, opened 2017-07-04
[Feature Request] ECDH Shared Secret scalarmult_*() API


psalami:
While it is shown in #61 that a wrapper for the ``crypto_kx_client_session_keys()`` and ``crypto_kx_server_session_keys()`` API is now implemented in the [sessionKeyPair](https://github.com/jedisct1/swift-sodium/blob/master/Sodium/KeyExchange.swift#L115) function, there is currently no wrapper available for the ``crypto_scalarmult_*()`` API. As noted in the [libsodium key exchange documentation](https://download.libsodium.org/libsodium/content/key_exchange/]) notes, the ``crypto_scalarmult_*()`` API should be used to build constructions that differ from the higher-level ``crypto_kx_*_session_keys()`` implementation.

See also the [libsodium Diffie-Hellman function documentation](https://download.libsodium.org/libsodium/content/advanced/scalar_multiplication.html).

Direct access to the X25519 function is necessary, for example, to implement a [double ratchet algorithm](https://matrix.org/git/olm/about/docs/olm.rst). In addition, access to the ``crypto_scalarmult_base()`` function is also necessary to derive a public key from a private key.

PR #61, opened 2017-03-17
Add key exchange

ChaosCoder:
This adds the key exchange functionality of libsodium to swift-sodium.

As I wanted to keep the structure of lib sodium, this PR introduces a new class KeyExchange with the following method:

```
func sessionKeyPair(publicKey: PublicKey, secretKey: SecretKey, otherPublicKey: PublicKey, side: Side) -> SessionKeyPair?
```

See https://github.com/jedisct1/swift-sodium/pull/60 and [libsodium.org/doc/key_exchange](https://download.libsodium.org/doc/key_exchange/)

PR #94, opened 2017-07-04
adding wrapper for Diffie-Hellman (ECDH) Shared Secret crypto_scalarmult_*() API

psalami:
To derive a shared secret given Alice's secret key and Bob's public key:
```swift
let sharedSecret = sodium.keyAgreement.sharedSecret(secretKey: aliceKeyPair.secretKey, publicKey: bobKeyPair.publicKey)
```

To derive a public key from a private key:
```swift
let aliceSecretKey = sodium.utils.hex2bin("a9029ec4ec56dd6f3ce5a5fa27a17a005ce73a5b8e77529887f24f73ffa10d67")!
let alicePublicKey = sodium.keyAgreement.publicKey(secretKey: aliceSecretKey)
```

Also adding unit tests and README entries to document the new functions.

see also #93

While it is shown in #61 that a wrapper for the ``crypto_kx_client_session_keys()`` and ``crypto_kx_server_session_keys()`` API is now implemented in the [sessionKeyPair](https://github.com/jedisct1/swift-sodium/blob/master/Sodium/KeyExchange.swift#L115) function, there is currently no wrapper available for the ``crypto_scalarmult_*()`` API. As noted in the [libsodium key exchange documentation](https://download.libsodium.org/libsodium/content/key_exchange/]) notes, the ``crypto_scalarmult_*()`` API should be used to build constructions that differ from the higher-level ``crypto_kx_*_session_keys()`` implementation.

See also the [libsodium Diffie-Hellman function documentation](https://download.libsodium.org/libsodium/content/advanced/scalar_multiplication.html).

Direct access to the X25519 function is necessary, for example, to implement a [double ratchet algorithm](https://matrix.org/git/olm/about/docs/olm.rst). In addition, access to the ``crypto_scalarmult_base()`` function is also necessary to derive a public key from a private key.


jedisct1:
Do you specifically need this?

The documentation you added is exactly why this API was not implemented, and why it never existed in libhydrogen. 

The output of the scalar multiplication should not be used as a shared secret. It must be hashed, along with information from all parties, typically public keys.

If you need a 256-bit key,`scalarmult()` is not a good fit as its range is smaller than the key space. Hashing the output works around this, adds some protection against weak public keys and and provides an authentication layer that may prevent vulnerabilities with badly designed protocols. 

`crypto_kx_` is usually the API you want to use instead. It produces two keys, but if you only need one, just pick one.

The raw output *could* be used with Olm as the aggregate keys are eventually hashed, but using the `crypto_kx_` output doesn't change its design or security properties.

PR #96, opened 2017-07-04
Fix allocated size of generic hash state

### LINKED ISSUE #92: GenericHash::Stream - testStreaming fails sometimes
SodiumTests.ReadmeTests sometimes (in about 30% of attempts) fails with the following error:

```
Test Case '-[SodiumTests.ReadmeTests testStreaming]' started.
xctest(6198,0x107d243c0) malloc: *** error for object 0x7fb14cd89ae8: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
```

Environment:
iPhone 7 Plus simulator, iOS 10.3

This could possibly be due to the [Stream.state](https://github.com/jedisct1/swift-sodium/blob/master/Sodium/GenericHash.swift#L144) property being an unsafe pointer.

blochberger:
As mentioned in the documentation, one should not rely on the value of `sizeof` but rather use `crypto_generichash_statebytes()`.

This fixes #92.

### DIFF
--- a/Sodium/GenericHash.swift
+++ b/Sodium/GenericHash.swift
@@ -144,7 +144,7 @@ public class GenericHash {
         private var state: UnsafeMutablePointer<crypto_generichash_state>?
 
         init?(key: Data?, outputLength: Int) {
-            state = UnsafeMutablePointer<crypto_generichash_state>.allocate(capacity: 1)
+            state = UnsafeMutablePointer<crypto_generichash_state>.allocate(capacity: crypto_generichash_statebytes())
             guard let state = state else {
                 return nil
             }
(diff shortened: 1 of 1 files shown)


jedisct1:
Saw your PR after having made a similar change.

The unit in `UnsafePointer.allocate()` is the given type, not a number of bytes, so I had to introduce some pointer conversions. Do you know of a cleaner way to achieve this?

blochberger:
Closed in favour of 1ed8ab39b274f1d5e1dd19779fae2d697000d299, which allocates the exact bytes required. Mine is allocating too much.

blochberger:
I think your way is the correct one.

Issue #97, opened 2017-07-10
[Feature Request] key derivation


psalami:
It appears that a wrapper for the ``crypto_kdf_derive_from_key()`` function does not yet exist. This function is necessary in order to derive subkeys from a master key, using the blake2b hash function. This is necessary, for example, to implement an OLM [double ratchet algorithm](https://matrix.org/git/olm/about/docs/olm.rst).

[libsodium key derivation function documentation](https://download.libsodium.org/doc/key_derivation/)

PR #98, opened 2017-07-10
adding wrapper for key derivation function

psalami:
To derive a subkey from an input (master) key:

```swift
let subKey1 = try sodium.keyDerivation.derive(fromKey: aliceKeyPair.secretKey, atIndex: 0, havingOutputBytes: 32, inContext: "TEST")
```

This function is a wrapper for the `crypto_kdf_derive_from_key()` function documented [here](https://download.libsodium.org/doc/key_derivation/).

Also added unit tests and README update.
This addresses #97.


jedisct1:
Merged!

Thanks a ton for including tests and having updated the documentation!

Sodium/Sodium.swift
```diff
@@ -21,6 +21,7 @@ public class Sodium {
     public let keyExchange = KeyExchange()
     public let auth = Auth()
     public let stream = Stream()
+	public let keyDerivation = KeyDerivation()
```
jedisct1: This pull request was almost perfect :)

psalami: I think we ran into the great question of tabs vs spaces. I have my Xcode settings set to tabs, which I believe is the default and also recommended by the [GitHub Swift Style Guide](https://github.com/github/swift-style-guide/blob/master/README.md#whitespace). ;-)

SodiumTests/SodiumTests.swift
```diff
@@ -284,4 +284,47 @@ class SodiumTests: XCTestCase {
         let verify2 = sodium.auth.verify(message: message, secretKey: key2, tag: tag)
         XCTAssertFalse(verify2)
     }
+	
+	func testKeyDerivationInputKeyTooShort() throws {
+		let sodium = Sodium()!
+		let seed = sodium.utils.hex2bin("00 11 22 33 44 55 66 77 88 99 aa bb cc dd ee ff 00 11 22 33 44 55 66 77 88 99 aa bb cc dd ee ff", ignore: " ")!
+		let inputKey = sodium.randomBytes.deterministic(length: 15, seed: seed)!
```
blochberger: I think deterministically generating random data is not required here. See also below.

SodiumTests/SodiumTests.swift
```diff
@@ -284,4 +284,47 @@ class SodiumTests: XCTestCase {
...
+		XCTAssertThrowsError(try sodium.keyDerivation.derive(fromKey: aliceSecretKey, atIndex: 0, inContext: "TEST_SODIUM"))
+	}
+	
+	func testKeyDerivation() throws {
+		let sodium = Sodium()!
+		let aliceSecretKey = sodium.utils.hex2bin("a9029ec4ec56dd6f3ce5a5fa27a17a005ce73a5b8e77529887f24f73ffa10d67")!
+		let subKey1 = try sodium.keyDerivation.derive(fromKey: aliceSecretKey, atIndex: 0, havingOutputBytes: 32, inContext: "TEST")
+		let subKey2 = try sodium.keyDerivation.derive(fromKey: aliceSecretKey, atIndex: 1, havingOutputBytes: 32, inContext: "TEST")
+		
+		NSLog(sodium.utils.bin2hex(subKey2)!)
```
blochberger: That probably is leftover debug code?

Sodium/KeyDerivation.swift
```diff
@@ -0,0 +1,74 @@
...
+	 - Returns: the derived key
+	 
+	 - Throws: NSError describing either an input value whose length does not fall within the specified bounds, or an error generated by the underlying libsodium function
+	
+	 - Note: Input and output keys must have a length between 16 and 64 bytes (inclusive), otherwise an error is thrown. Context must be at most 8 characters long. If the specified context is shorter than 8 characters, it will be padded to 8 characters.
+	
+	 */
+	public func derive(fromKey inputKey: Key, atIndex index: UInt64, havingOutputBytes outputBytes: Int = crypto_kdf_keybytes(), inContext context: String) throws -> Key {
+		if outputBytes < crypto_kdf_bytes_min() {
+			throw NSError(domain: "libsodium", code: -10, userInfo: [NSLocalizedDescriptionKey: String(format:"the length of the derived key must be at least %d bytes", crypto_kdf_bytes_min())])
```
blochberger: The rest of the project returns `nil` on error. This introduces differences in error handling throughout the project, which makes the API hard to use.

I think making errors more explicit than simply returning `nil` is the better approach in general. Instead of using `NSError` we could leverage the `Error` protocol and return type-safe errors. I would especially avoid hardcoded error values like `-10`.

Sodium/KeyDerivation.swift
```diff
@@ -0,0 +1,74 @@
...
+		}
+		
+		if outputBytes > crypto_kdf_bytes_max() {
+			throw NSError(domain: "libsodium", code: -11, userInfo: [NSLocalizedDescriptionKey: String(format:"the length of the derived key must be less than or equal to %d bytes", crypto_kdf_bytes_max())])
+		}
+		
+		if inputKey.count < crypto_kdf_bytes_min() {
```
blochberger: The `inputKey` should be of size `crypto_kdf_keybytes()`.

SodiumTests/SodiumTests.swift
```diff
@@ -284,4 +284,47 @@ class SodiumTests: XCTestCase {
...
+	func testKeyDerivationContextTooLong() throws {
+		let sodium = Sodium()!
+		let aliceSecretKey = sodium.utils.hex2bin("a9029ec4ec56dd6f3ce5a5fa27a17a005ce73a5b8e77529887f24f73ffa10d67")!
+		XCTAssertThrowsError(try sodium.keyDerivation.derive(fromKey: aliceSecretKey, atIndex: 0, inContext: "TEST_SODIUM"))
+	}
+	
+	func testKeyDerivation() throws {
+		let sodium = Sodium()!
+		let aliceSecretKey = sodium.utils.hex2bin("a9029ec4ec56dd6f3ce5a5fa27a17a005ce73a5b8e77529887f24f73ffa10d67")!
+		let subKey1 = try sodium.keyDerivation.derive(fromKey: aliceSecretKey, atIndex: 0, havingOutputBytes: 32, inContext: "TEST")
```
blochberger: I would add an additional test case, where deriving the same key twice leads to the same sub key. Additionally, you perform some padding on the context string, which is untested.

Sodium/KeyDerivation.swift
```diff
@@ -0,0 +1,74 @@
...
+		
+		if inputKey.count > crypto_kdf_bytes_max() {
+			throw NSError(domain: "libsodium", code: -13, userInfo: [NSLocalizedDescriptionKey: String(format:"the length of the input key must be less than or equal to %d bytes", crypto_kdf_bytes_max())])
+		}
+		
+		if context.lengthOfBytes(using: .utf8) > crypto_kdf_contextbytes() {
+			throw NSError(domain: "libsodium", code: -14, userInfo: [NSLocalizedDescriptionKey: String(format:"the length of the context String must be less than or equal to %d bytes", crypto_kdf_contextbytes())])
+		}
+		
+		let contextPadded = context.padding(toLength: crypto_kdf_contextbytes(), withPad: " ", startingAt: 0)
```
blochberger: I'm unsure if implicitly padding the context is a good thing – but at least it is mentioned in the API documentation ;)

PR #100, opened 2017-07-14
Cleanup and improve recent tests

blochberger:
This makes some changes to the tests that were added lately:

- Hard-coded numbers are avoided
- Key derivation test is now based on logic rather than specific values
- Some code cleanup

### DIFF
--- a/SodiumTests/SodiumTests.swift
+++ b/SodiumTests/SodiumTests.swift
@@ -286,8 +286,7 @@ class SodiumTests: XCTestCase {
     }
 
     func testKeyDerivationInputKeyTooShort() {
-        let seed = sodium.utils.hex2bin("00 11 22 33 44 55 66 77 88 99 aa bb cc dd ee ff 00 11 22 33 44 55 66 77 88 99 aa bb cc dd ee ff", ignore: " ")!
-        let secretKey = sodium.randomBytes.deterministic(length: sodium.keyDerivation.KeyBytes - 1, seed: seed)!
+        let secretKey = sodium.randomBytes.buf(length: sodium.keyDerivation.KeyBytes - 1)!
 
         XCTAssertNil(sodium.keyDerivation.derive(secretKey: secretKey, index: 0, length: sodium.keyDerivation.BytesMin, context: "TEST"))
     }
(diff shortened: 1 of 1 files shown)


jedisct1:
Thanks!

Removing the barcoded numbers is good, but I'd like to keep some deterministic tests. If only to make sure that we are properly passing the arguments to the native function.

blochberger:
Are there any official test vectors for `libsodium`? Otherwise there is no point in adding the values you get when simply calling the Swift version of the function.

jedisct1:
They aren't, but you can obtain the value by calling the C function directly, without the Swift wrappers.

blochberger:
That's quite cumbersome. You have to trust the author of a change, that he has actually taken the values from invoking the C functions.

We could link `libsodium` against the test directly and compare the result of calls to `libsodium` and the Swift wrappers. But then you might make a mistake invoking the C function within the test case as well.

We could also include the `*.exp` files from [jedisct1/libsodium](https://github.com/jedisct1/libsodium) and reproduce it within the Swift test cases.

Any other ideas? In my opinion official test vectors would be the best way to go. The `*.exp` files are somewhat going in that direction already.

jedisct1:
https://github.com/jedisct1/libsodium-validation 's goal was to solve this, but I haven't found the time to keep working on it.

jedisct1:
Even a single deterministic test case is better than none, if only to spot regressions.

blochberger:
Okay, I added a regression test and fixed funky indentation (those tabs…). I keep the commit 843fe47b8a8cf6681f6b40bbe22d17cf75192b07 because the input value does not need to be deterministic if the actual value of the output is not taken into account.

jedisct1:
Thanks, Max!

blochberger:
Did you change your strategy on merging pull requests? It looks like they where all squashed into a single commit. This makes reverts harder and removes signatures from commits and I always have to change my local history.

jedisct1:
That's now the default when merging from the Github web interface :/

blochberger:
Strange decision changing that. Maybe disabling squash and rebase merging strategies in the projects settings helps?

jedisct1:
Disabled, thanks!

blochberger:
Looks like that setting helped. The commit in #102 was merged as expected.

PR #104, opened 2017-09-05
Migrate code for Xcode 9 beta 6

ziogaschr:


Sodium/RandomBytes.swift
```diff
@@ -57,7 +57,7 @@ public class RandomBytes {
      - Returns: The generated data.
      */
     public func deterministic(length: Int, seed: Data) -> Data? {
-        if length < 0 || seed.count != SeedBytes || length > 0x4000000000 {
+        if length < 0 || seed.count != SeedBytes || length > UInt.max {
```
ziogaschr: @jedisct1 can you please cross check that this change is correct? The compiler was giving an issue about what will happen on 32bit systems.

jedisct1: The limit should really be `0x4000000000`. The function call will crash with a higher value, which can happen if `UInt.max` is 64 bits.

ziogaschr: I reverted back this change and compiler gave no issues. It might be a compile issue during the conversion. Sorry for the false alarm and thanks for pointing the issue with regards the change.

ziogaschr:
I just saw you did the migration yourself. Thanks for being so fast @jedisct1. 👍

PR #106, opened 2017-09-06
Improve Sodium's initializer

zetasq:
@jedisct1 Since the original initializer calls **abort()** when **sodium_init()** returns -1, we can make the initializer non-optional.


zetasq:
Sorry, it seems the issue is already explained.

blochberger:
For reference, see #54.

PR #122, opened 2018-01-16
Fix warning

blochberger:
When building the "SodiumTests" target the following warning was issued:

> `ld: warning: directory not found for option '-F/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator11.2.sdk/Developer/Library/Frameworks'`

I validated that there is no such path in my Xcode installation and removed the framework search path accordingly. Since building succeeds anyway it looks like this is not required.


jedisct1:
Sure!

Thanks again, Max!

PR #124, opened 2018-02-22
Add Swift Package Manager and Linux support

### LINKED ISSUE #69: [Feature Request] Add SPM support
Add support for swift package manager please :D

tiwoc:
This PR adds SPM and Linux support, along with a bugfix (the macOS applicationDidFinishLaunching signature didn't work with newer versions of Swift).

Tested on Ubuntu 16.04 with Swift 4.0.3 from swift.org and libsodium 1.0.16 built with

```sh
./configure --prefix /usr
make
make check
sudo make install
```

Fixes #69


jedisct1:
Thanks!

But how does that work? Does it still use the precompiled libraries?

tiwoc:
SPM doesn't support precompiled libraries, unfortunately.

I've chosen the canonical way of using C libraries in SPM: Building a wrapper module that uses the library and headers as found on the system.

It should theoretically be possible to build libsodium from source as part of the SPM build, but one has absolutely no control over the compiler flags, so this would not be a great choice.

jedisct1:
This is not great, as system packages, especially on Linux, as often out of date.

The current precompiled binaries contain optimized versions for every ARM variant found on iOS devices. This cannot be something installed on the system.

tiwoc:
You're right, system packages might be out of date. SPM finds headers and libraries by loading the `pkg-config` file of the library, `libsodium.pc` in this case. One can use one's own libsodium build instead of the one provided by the system by setting the `PKG_CONFIG_PATH` environment variable before invoking `swift build` so that SPM finds the .pc file of the custom build. This way, one isn't forced to use the system-provided build.

I'd be happy to add a section about that to the README if you're generally interested in this PR.

Building for iOS is another task that SPM currently doesn't support, so this should - at least now - not be a problem. SPM is usually used to build server-side Swift apps. When SPM supports iOS at some point in the future, it will probably look for headers and libraries in a target-specific path and not in the host system, just like xcodebuild does when building for iOS. Alternatively, one will probably be able to use the `PKG_CONFIG_PATH` method outlined above.

jedisct1:
Merged, thanks!

jedisct1:
Reverted :( It breaks the iOS builds :(

tiwoc:
I'm really sorry. What is it that did break? I'm quite sure to have run the tests and demo apps.

jedisct1:
Absolutely nothing, sorry. This completely unrelated, and due to missing bitcode in object files.
Your changes have been merged back. Sorry for the false alarm, and thanks for your contribution!

blochberger:
Is there a better way to handle the changes introduced in commit 9442d45? That kind of thing tends to get out of sync.

jedisct1:
Indeed, I'm not a big fan of these changes. Whenever tests are removed or added, it is very likely that these lists will be forgotten.

tiwoc:
This is an area where Swift on Linux is unfinished. XCTest's autodetection of tests relies on the Objective-C runtime which is not available on Linux. The Swift runtime does not provide adequate support for introspection: https://bugs.swift.org/browse/SR-710

The workaround that I use in my server-side Swift projects is a [Sourcery template](https://oleb.net/blog/2017/03/keeping-xctest-in-sync/) (see "Appendix: Code generation with Sourcery" at the bottom), so that a LinuxMain.swift and the allTests arrays are automatically generated.

PR #125, opened 2018-02-23
Return non-optional key if keygen is guaranteed to succeed

tiwoc:
Returning a non-optional value saves the client code from unnecessary
error checking or generous use of force-unwraps.


blochberger:
Maybe the discussion in #54 applies here as well. The real question is: Might key generation functions fail, e.g., in future changes?

jedisct1:
I don't think they will ever fail. As long as `sodium_init()` passed, they will succeed.

PR #126, opened 2018-02-24
Revert "Add Swift Package Manager and Linux support"

jedisct1:
Reverts jedisct1/swift-sodium#124


jedisct1:
My bad, looks like this is an unrelated issue with bitcode.

PR #131, opened 2018-03-22
Aead Api

umbri:
based on https://github.com/aidantwoods/swift-paseto/commit/f3bc5c9852d86b56fdfb5caca09f5206504d7dbf

I think it make send to add methods like in https://github.com/jedisct1/swift-sodium/blob/master/Sodium/SecretBox.swift#L47


jedisct1:
Thanks!

Just two things:

- `additionalData` should probably be optional. Maybe by providing additional functions that don't require it, similar to what is being done in other classes.
- Would it be possible to update the doc as well?

Thanks!

umbri:
Sure, I am doing it now, and I will add ```detached``` mode too

aidantwoods:
Thanks for adding this in @umbri :)

PR #135, opened 2018-04-18
Return non-optional key if keygen is guaranteed to succeed (with docs)

tiwoc:
Returning a non-optional value saves the client code from unnecessary
error checking or generous use of force-unwraps.

This is a revised version of #125 which got reverted in #129 due to the
README not being updated.


tiwoc:
I'm sorry that this lingered for so long. I hadn't noticed #125 being reverted, but now my notification settings are updated.

PR #139, opened 2018-04-23
Add barebones .travis.yml

aidantwoods:
Since you use Swift package manager, running the unit tests in CI is pretty easy. If this is desired, you'll of course need to configure Travis to actually test this repo too :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,21 @@
+os: osx
+language: objective-c
+
+cache:
+  directories:
+    - $HOME/Library/Caches/Homebrew
+
+matrix:
+  include:
+    - osx_image: xcode9.3
+    - osx_image: xcode9.2
+    - osx_image: xcode9.1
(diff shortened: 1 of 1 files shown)


aidantwoods:
If you want to see an example Travis build, this file is basically just ripped from one of my projects – so a build will look a little like: https://travis-ci.org/aidantwoods/swift-paseto/builds/369824524.

You could also shave a considerable amount of time off the build depending on how pedantic you are about Homebrew being up to date before installing `libsodium`.

aidantwoods:
@jedisct1 any chance at getting Travis set up? :) While I run the tests locally on the latest installation of Swift, this is probably the easiest way of verifying source compatibility with Swift 4.0 (I don't have it to test on)

blochberger:
Can we enable testing for iOS in Travis as well?

blochberger:
I missed that, thanks!

PR #140, opened 2018-04-23
State stream protocol

aidantwoods:
There are a few classes (`GenericHash.Stream`, `SecretStream.XChaCha20Poly1305.PushStream`, and `SecretStream.XChaCha20Poly1305.PullStream`) which have some very similar internal behaviour which boils down to picking a types and specifying a value.

Using a new `StateStream` protocol we can allow these implementations to just specify this information and have `free()` and state generation be taken care of by a protocol extension.

I've also removed the optional wrapper around `state` since 1. the `.some` state was forced unwrapped in almost all of its use (either explicitly with `!` or implicitly via use of the C API); 2. all calls to `free()` were made right before the `StateStream` object was deinitialised or failed to initialise – either way – there would be no subsequent direct access to the pointer variable `state` because there would be no way to refer to the variable, 3. there appear to be no point during which normal function of these objects would permit the `nil` state (it was only set to this state before the object is thrown away). Thus it seemed that setting `state` to `nil` was both unnecessary and the `nil` state was not taken advantage of for preventing use of `nil` as a pointer (due to the forced unwraps).

### DIFF
--- a/Sodium/GenericHash.swift
+++ b/Sodium/GenericHash.swift
@@ -131,7 +131,7 @@ public class GenericHash {
         public var outputLength: Int = 0
 
         init?(key: Data?, outputLength: Int) {
-            state = Stream.gen(capacity: capacity)
+            state = Stream.generate()
             var result: Int32 = -1
             if let key = key {
                 result = key.withUnsafeBytes { keyPtr in
--- a/Sodium/GenericHash.swift
+++ b/Sodium/GenericHash.swift
@@ -125,7 +125,7 @@ public class GenericHash {
 
     public class Stream: StateStream {
         typealias State = crypto_generichash_state
-        var capacity = crypto_generichash_statebytes()
+        static var capacity = crypto_generichash_statebytes()
         var state: UnsafeMutablePointer<State>
 
         public var outputLength: Int = 0
--- a/Sodium/GenericHash.swift
+++ b/Sodium/GenericHash.swift
@@ -126,7 +126,7 @@ public class GenericHash {
     public class Stream: StateStream {
         typealias State = crypto_generichash_state
         static var capacity = crypto_generichash_statebytes()
-        var state: UnsafeMutablePointer<State>
+        private var state: UnsafeMutablePointer<State>
 
         public var outputLength: Int = 0
 
@@ -145,6 +145,10 @@ public class GenericHash {
                 return nil
             }
             self.outputLength = outputLength
+        }
+
+        private func free() {
+            Stream.free(state)
         }
 
         deinit {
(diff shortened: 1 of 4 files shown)


Sodium/GenericHash.swift
```diff
@@ -147,6 +147,10 @@ public class GenericHash {
             self.outputLength = outputLength
         }
 
+        private func free() {
+            Stream.free(state)
```
tiwoc: In its current form, `StateStream` can be implemented privately (inside a `private extension Stream: StateStream`), so that all the members that it prescribes could be private without the workaround of implementing a wrapper `free()`. If done this way, the changes of this commit are unnecessary and the code would IMHO be clearer.

Sodium/GenericHash.swift
```diff
@@ -123,16 +123,15 @@ public class GenericHash {
         return Stream(key: nil, outputLength: outputLength)
     }
 
-    public class Stream {
+    public class Stream: StateStream {
+        typealias State = crypto_generichash_state
+        static var capacity = crypto_generichash_statebytes()
```
tiwoc: This should be a `let`, it doesn't (and mustn't) change.

Sodium/SecretStream.swift
```diff
@@ -60,19 +60,19 @@ public class SecretStream {
             return stream
         }
 
-        public class PushStream {
-            private var state: UnsafeMutablePointer<crypto_secretstream_xchacha20poly1305_state>?
+        public class PushStream: StateStream {
+            typealias State = crypto_secretstream_xchacha20poly1305_state
+
+            static var capacity = crypto_secretstream_xchacha20poly1305_statebytes()
```
tiwoc: ditto

Sodium/SecretStream.swift
```diff
@@ -128,31 +128,25 @@ public class SecretStream {
...
                 free()
             }
         }
 
-        public class PullStream {
-            private var state: UnsafeMutablePointer<crypto_secretstream_xchacha20poly1305_state>?
+        public class PullStream: StateStream {
+            typealias State = crypto_secretstream_xchacha20poly1305_state
+
+            static var capacity = crypto_secretstream_xchacha20poly1305_statebytes()
```
tiwoc: ditto

Sodium/GenericHash.swift
aidantwoods: I don't think that declaring an accessibility level on conformance is possible – doesn't it have to be the same visibility as the protocol?

tiwoc: You're right, I just tried it out. This does only work for extensions that don't declare a protocol conformance.

tiwoc:
LGTM

PR #141, opened 2018-04-26
Tweaks for readability

aidantwoods:


aidantwoods:
The last commit introduces a new type
```swift
enum ExitCode {
    case SUCCESS
    case FAILURE
}
```

with the following constructor:
```swift
init (from int: Int32) {
    switch int {
    case 0:  self = .SUCCESS
    default: self = .FAILURE
    }
}
```

If it's ever desired, more exit code states could be added. By making use of Swift's exhaustive switch-cases, this will provide added safety that all exit codes are considered where we are trying to enumerate over all of them

As for the benefits right now, IMO `result == .SUCCESS` is a clearer statement than `result == 0`.

The commit also includes an extension to `Int32` which adds the computed property `exitCode: ExitCode` to map itself into an `ExitCode` state. This allows us to retrieve an `ExitCode` from a function that returns a `Int32` by writing something like `crypto_box_keypair(pkPtr, skPtr).exitCode`.

PR #144, opened 2018-04-27
Defer to Swift's own handling instead of manually using pointers

aidantwoods:
This is based on #142, so commits prior to 48b08e4 should be ignored (actual diff: https://github.com/aidantwoods/swift-sodium/compare/enhancement/protocol-abstractions...aidantwoods:readability/less-explict-pointers).

Swift provides native handling for "coercing" pointers out of standard types when interacting with C functions, this is documented in [Interacting with C Pointers](https://developer.apple.com/swift/blog/?id=6).

Essentially this allows a C function that is written to accept something like `UnsafePointer<UInt8>` to instead accept `[UInt8]` (`Array<UInt8>`).  For something that would accept a mutable pointer, we only need prepend an `&` to the variable (as we would in normal Swift code for mutable arguments). This is mostly just syntactic sugar for making those nested `.withUnsafePointer` style calls more readable.

We can't immediately use C functions though, because we work with `Data` most of the time (which doesn't have an explicit type that Swift can make use of). To transition: I've introduced a new typealias `Bytes = Array<UInt8>`, as well as adding an extension to `Array<UInt8>` so that we can initialise the size with zeroed bytes (this is identical to the behaviour of `Data(count:)`). For variables that require space prior to calling a C function: i.e. functions that return messages, signatures, etc...  essentially all that is required is renaming `Data(count:)` to `Bytes(count:)`.

As mentioned in the Swift post, it isn't possible to grow the size of byte arrays that are converted to pointers using this method, so for things like padding, the `Data` type is still used so that the size can be increased.

### Public API changes
IMO it would be a good idea to just accept byte arrays instead of `Data` in the public interface, so I've changed everything to do that – this additionally motivated by avoiding needing to do conversions from `Data` to `Bytes` inside the function bodies.
`Bytes` (`Array<UInt8>`) is also quite easy to construct as a user, and things like `string.utf8` provide *almost* bytes anyway (I've added an extension to `String` to just convert straight to bytes too). This of course can be reverted if there is desire to keep the public API the same though.


aidantwoods:
Just to highlight an example of one of the diffs achieved here by using Swift's native handling :)

```diff
-            let result: ExitCode
 
-            if let additionalData = additionalData {
-                result = authenticatedCipherText.withUnsafeMutableBytes { authenticatedCipherTextPtr in
-                    authenticatedCipherTextLen.withUnsafeMutableBytes { authenticatedCipherTextLenPtr in
-                        message.withUnsafeBytes { messagePtr in
-                            additionalData.withUnsafeBytes { additionalDataPtr in
-                                nonce.withUnsafeBytes { noncePtr in
-                                    secretKey.withUnsafeBytes { secretKeyPtr in
-                                        crypto_aead_xchacha20poly1305_ietf_encrypt(
-                                            authenticatedCipherTextPtr,
-                                            authenticatedCipherTextLenPtr,
-
-                                            messagePtr,
-                                            UInt64(message.count),
-
-                                            additionalDataPtr,
-                                            UInt64(additionalData.count),
-
-                                            nil, noncePtr, secretKeyPtr
-                                        ).exitCode
-                                    }
-                                }
-                            }
-                        }
-                    }
-                }
-            } else {
-                result = authenticatedCipherText.withUnsafeMutableBytes { authenticatedCipherTextPtr in
-                    authenticatedCipherTextLen.withUnsafeMutableBytes { authenticatedCipherTextLenPtr in
-                        message.withUnsafeBytes { messagePtr in
-                            nonce.withUnsafeBytes { noncePtr in
-                                secretKey.withUnsafeBytes { secretKeyPtr in
-                                    crypto_aead_xchacha20poly1305_ietf_encrypt(
-                                        authenticatedCipherTextPtr,
-                                        authenticatedCipherTextLenPtr,
-
-                                        messagePtr,
-                                        UInt64(message.count),
-
-                                        nil,
-                                        0,
-
-                                        nil, noncePtr, secretKeyPtr
-                                    ).exitCode
-                                }
-                            }
-                        }
-                    }
-                }
-            }
-            guard result == .SUCCESS else { return nil }
+            guard .SUCCESS == crypto_aead_xchacha20poly1305_ietf_encrypt (
+                &authenticatedCipherText, &authenticatedCipherTextLen,
+                message, UInt64(message.count),
+                additionalData, UInt64(additionalData?.count ?? 0),
+                nil, nonce, secretKey
+            ).exitCode else { return nil }
```

jedisct1:
Neat!

Sodium/Bytes.swift
```diff
@@ -0,0 +1,21 @@
...
+        return String(data: Data(bytes: self), encoding: .utf8)
+    }
+}
+
+extension ArraySlice where Element == UInt8 {
+    var bytes: Bytes { return Bytes(self) }
+}
+
+public extension String {
+    var bytes: Bytes { return Bytes(self.utf8) }
```
blochberger: I would make it more clear that the string is UTF-8 encoded by calling this `utf8bytes`.

Sodium/ExitCode.swift
```diff
@@ -0,0 +1,15 @@
+enum ExitCode {
```
blochberger: This is never used as an actual exit code, so I think the name might be slightly confusing. I suggest to rename it to status something.

In addition, each function might behave differently, e.g., `sodium_init()` can return `1`, which indicates that the library has already been initialized, but is not a failure technically. Therefore, I think a generic approach is more confusing than helpful. For syntactic sugaring we could use `noErr` or define a `SUCCESS` constant.

aidantwoods: The idea behind the name was to make implicit the general use-case that the integer `0` would correspond to the success state.

I take your point about it not being so easily useful when functions aren't as simple as `0` for success, and failure otherwise though.

We could perhaps add more states, but I'm now questioning the value add altogether. When this was written, at the use site there were lots of nested closures surrounding the function due to explicit pointer use (and so IMO this helped highlight some of the intent upfront because the actual function call could be something like five layers of nesting deep). This isn't really a concern any more, and so perhaps this is no longer justified since the function call is far more readable in the first place.

PR #148, opened 2018-04-27
Remove stream pointers

aidantwoods:
This is based on #147,  so please ignore commits prior to dcbc8a5. ([actual diff](https://github.com/aidantwoods/swift-sodium/compare/enchancement/types...aidantwoods:enhancement/remove-stream-pointers))

This directly instantiates stream states and forgoes using a pointer until interacting with the C API again (at which point we use Swift's native handling).

Because we're no longer holding onto a pointer, I don't believe there is any reason to need to proactively deallocate the state (as was done when using a pointer to the state). Instead the state should just be deallocated when there type containing it is no longer referenced.

Provided that the above is sane, this means that this type no longer needs to be a class and can instead be a struct – which not only allows us to declare functions as mutating, but also requires it 
 because it's now a value type. This also means the user must create streams as a `var` when they want to update them after init (which is probably a little more honest than a `let` :) )

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ let secretkey = sodium.secretStream.xchacha20poly1305.key()
 
 /* stream encryption */
 
-let stream_enc = sodium.secretStream.xchacha20poly1305.initPush(secretKey: secretkey)!
+var stream_enc = sodium.secretStream.xchacha20poly1305.initPush(secretKey: secretkey)!
 let header = stream_enc.header()
 let encrypted1 = stream_enc.push(message: message1)!
 let encrypted2 = stream_enc.push(message: message2)!
@@ -59,7 +59,7 @@ let encrypted3 = stream_enc.push(message: message3, tag: .FINAL)!
 
 /* stream decryption */
 
-let stream_dec = sodium.secretStream.xchacha20poly1305.initPull(secretKey: secretkey, header: header)!
+var stream_dec = sodium.secretStream.xchacha20poly1305.initPull(secretKey: secretkey, header: header)!
 let (message1_dec, tag1) = stream_dec.pull(cipherText: encrypted1)!
 let (message2_dec, tag2) = stream_dec.pull(cipherText: encrypted2)!
 let (message3_dec, tag3) = stream_dec.pull(cipherText: encrypted3)!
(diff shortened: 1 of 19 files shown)


aidantwoods:
#### TL;DR for the below
Should a user copy a push (respectively pull, ...) stream, do we want both copies to:
1. be aliases of each other, referring to a single linear shared state; or
2. individually behave as independent (potentially diverging) states after being copied.

---

Something worth pointing out here is that by making streams value types we inherit properties of value types – specifically copies do not share writeable references to the same data in memory.

This means that if someone writes something like:

```swift
let sodium = Sodium()
let message_1 = "My favourite colour is: ".bytes
let message_a2 = "blue!".bytes
let message_b2 = "pink!".bytes

let secretkey = sodium.secretStream.xchacha20poly1305.key()

var stream_a = sodium.secretStream.xchacha20poly1305.initPush(secretKey: secretkey)!
let header = stream_a.header()

let encrypted_1 = stream_a.push(message: message_1)!

var stream_b = stream_a

let encrypted_a2 = stream_a.push(message: message_a2, tag: .FINAL)!
let encrypted_b2 = stream_b.push(message: message_b2, tag: .FINAL)!
```

Then they will have essentially forked the steam where these two streams would share the first encrypted message `encrypted_1`, but their final messages would differ (i.e. given `encrypted_1` either `encrypted_a2` or `encrypted_b2` could follow).

IMO this is unsurprising behaviour given the above code (Swift has value-types specifically to prevent unintended/accidental implicit sharing), simultaneously I suspect having multiple *actual copies* of a stream is unlikely to be desired, or even correct.

If we do not want to permit such behaviour at all then we should retain streams being classes so that they are reference types and they share data if copied.  Reference-type behaviour is also the current behaviour (which may be an argument for not changing it in itself).

Ultimately this just comes down to what is the most desired behaviour should a user write `var stream_b = stream_a` (or equivalent, e.g. passing a stream instance into a function as a non-mutable param (i.e. no `inout`)).

Regardless of whether or not we use value or reference types here, we can still get rid of the the explicit use of pointers (which was really the goal here).

tiwoc:
I think it would be surprising for most users of the library if an assignment of a stream would create a fork of the stream (i.e. when the stream was a value type). I'm also not sure whether this would even be secure, because creation of the stream might use some nonce which would be reused if the state was copied and reused for another stream.

Therefore, I'd prefer keeping the reference type.

aidantwoods:
37d23ff reverts making these value types.

PR #149, opened 2018-04-28
[WIP] BytesRepresentable protocol

aidantwoods:
Recall #144 changed some functions to accept `Bytes` instead of `Data`.

This PR introduces a `BytesRepresentable` protocol that types can implement to specify how they should be converted to bytes and constructed from bytes. This protocol requires the conforming type to always succeed in converting to `Bytes`, but allows for the possibility of failure the other way round.

Any conforming type will also receive a default implementation for constructing itself from an arbitrary `BytesRepresentable`.

Included also are conformance implementations for `String` `[UInt8]` (`Bytes`), `ArraySlice<UInt8>`, and `Data`.

This means that

```swift
let encryptedMessageToBob =
    sodium.box.seal(message: message, recipientPublicKey: bobKeyPair.publicKey)!
```

is valid code when `message` is any of `String`, `Bytes`, `Data`, or any other `BytesRepresentable`.

---

This is based on #148, so the commits relevant only to this PR begin at 0ee6d72 ([actual diff](https://github.com/aidantwoods/swift-sodium/compare/enhancement/remove-stream-pointers...aidantwoods:enhancement/bytes-representable-protocol))

### DIFF
--- a/Sodium/BytesRepresentables/Array+Extension.swift
+++ b/Sodium/BytesRepresentables/Array+Extension.swift
@@ -1,0 +1,7 @@
+extension Array: BytesRepresentable where Element == UInt8 {
+    public var bytes: [UInt8] { return self }
+
+    public init (bytes: [UInt8]) {
+        self = bytes
+    }
+}
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ let secretkey = sodium.secretStream.xchacha20poly1305.key()
 
 /* stream encryption */
 
-let stream_enc = sodium.secretStream.xchacha20poly1305.initPush(secretKey: secretkey)!
+var stream_enc = sodium.secretStream.xchacha20poly1305.initPush(secretKey: secretkey)!
 let header = stream_enc.header()
 let encrypted1 = stream_enc.push(message: message1)!
 let encrypted2 = stream_enc.push(message: message2)!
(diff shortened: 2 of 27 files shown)


Sodium/BytesRepresentables/Array+Extension.swift
```diff
@@ -0,0 +1,7 @@
+extension Array: BytesRepresentable where Element == UInt8 {
+    public var bytes: [UInt8] { return self }
+
```
aidantwoods: Gah, this might be a non-starter – I forgot already that Swift 4.0 doesn't support conditional conformance:
```
error: extension of type 'Array' with constraints cannot have an inheritance clause
extension Array: BytesRepresentable where Element == UInt8 {
^                ~~~~~~~~~~~~~~~~~~
```

aidantwoods: This could potentially be solved by making `Bytes` a stand-alone type (I.e. a struct that contains a [UInt8] so that `Bytes` can conform to `BytesRepresentable`. The conformances that require a `where` condition can be conditionally enabled in Swift >= 4.1, and Swift 4.0 can still use a constructor to easily create `Bytes` from these specialised generic types. 

Problematically if `Bytes` is a stand-alone type, then we need to further convert to `[UInt8]` before use with the C functions (because these require a `[UInt8]` for the pointer coercion), so perhaps `BytesContainer` would be a better name instead of `Bytes`.

I'm not a major fan of this solution. Anyone got any other ideas here? :)

aidantwoods:
Closing since I'm not really sure there is an acceptable way to implement this pre Swift 4.1.

PR #150, opened 2018-05-04
BytesRepresentable protocol

aidantwoods:
Recall #144 changed most functions to accept `Bytes` instead of `Data`. Essentially this PR aims to turn some of these changes into expansions: making it easier to use common types like `String`, or `Data` with the API and have it "just work" because there are very reasonable general purpose methods of converting these into bytes.

These changes also allow a user specify how to convert any of their own types to and from bytes. Should they do this they will be able to use their own types directly with the API without explicit conversion too.

This PR introduces a `BytesRepresentable` protocol that types can implement to specify how they should be converted to bytes and constructed from bytes. This protocol requires the conforming type to always succeed in converting to bytes, but allows for the possibility of failure the other way round.

Any conforming type will also receive a default implementation for constructing itself from an arbitrary `BytesRepresentable`.

Because most API functions can now be written to accept any `BytesRepresentable`, previous code which passed `Data` need not convert any of its arguments.

Included also are conformance implementations for `String`, `Data`, and `BytesContainer`. Additionally conformances for `[UInt8]` and `ArraySlice<UInt8>` are available on Swift >= 4.1 only.

This means that something like

```swift
let encryptedMessageToBob =
    sodium.box.seal(message: message, recipientPublicKey: bobKeyPair.publicKey)!
```

is valid code when `message` is any of `String`, `Data`, or any other `BytesRepresentable`.

(This is additional work on top of the closed #149 – new PR because I forced pushed to the branch to clean up some merge commits, which apparently blocks reopening):

A `BytesContainer` wrapper type was added (to wrap the generic `[UInt8]` which was previously called `Bytes` by alias – the wrapper is added in place of this alias). Swift 4.0 doesn't permit conforming on specialised generics, which is a bit of a problem if we use `Bytes` (a specialised generic) as a common-currency type, because it won't be able to conform the `BytesRepresentable` protocol (the whole purpose of the protocol is to encompass things that can be represented as `Bytes`).

To get around this `BytesContainer` would become the common-currency – which *can* conform to `BytesRepresentable` because it is not generic (and so doesn't require conditional conformance from Swift 4.1).
Most functions will now accept a `BytesRepresentable`, and will return in terms of `BytesContainer`. It is therefore desirable that these `BytesContainer`s conform to `BytesRepresentable` so that they can be passed back into other parts of the API (this was not possible in Swift 4.0 with the `Bytes` typealias).

---

This is based on #148, so the commits relevant only to this PR begin at 0ee6d72 ([actual diff](https://github.com/aidantwoods/swift-sodium/compare/enhancement/remove-stream-pointers...aidantwoods:enhancement/bytes-representable-protocol))


aidantwoods:
As a TL;DR, the goal here is probably best summed up by looking at the [diff to README.md](https://github.com/aidantwoods/swift-sodium/compare/enhancement/remove-stream-pointers...aidantwoods:enhancement/bytes-representable-protocol#diff-04c6e90faac2675aa89e2176d2eec7d8)

aidantwoods:
@jedisct1 if there is appetite to go down this route I can resolve the conflicts?

Sodium/Utils.swift
```diff
@@ -16,11 +16,23 @@ extension Utils {
...
+    }
+
+    /**
+     Tries to effectively zero bytes in `data`, even if optimizations are being applied to the code.
+
+     - Parameter data: The `BytesContainer` object to zero.
+     */
+    public func zero(_ data: inout BytesContainer)  {
+        let count = data.count
+        sodium_memzero(&data.bytes, count)
```
blochberger: Does this make sense? I am not sure, how bytes are handled internally but I can imagine that `BytesRepresentable.bytes` is implemented in a way, that the bytes are copied. The implicit copy created will denn be zeroed but the original value will stay in place. The user of this API might think that something has been securely erased, when it actually was not. Correct me if I am wrong.

aidantwoods: Let me know if I'm off here, but did you perhaps misread the type (you mention `BytesRepresentable` here)? `data` is a `BytesContainer` (i.e. a struct which wraps the stored property `var bytes: [UInt8]`), but not a general `BytesRepresentable`. Thus this method applies to a known implementation, and the zeroing should be directly applied to the stored property `bytes` by reference?

For sure this couldn't be implemented over a general `BytesRepresentable` for the reasons above though.

README.md
```diff
@@ -303,24 +303,24 @@ let subKey2 = sodium.keyDerivation.derive(secretKey: secretKey,
...
 sodium.utils.zero(&dataToZero)
 ```
 
 ### Constant-time comparison
 
 ```swift
 let sodium = Sodium()
-let secret1 = "Secret key".bytes
-let secret2 = "Secret key".bytes
+let secret1 = BytesContainer("Secret key")
```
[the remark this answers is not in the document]
aidantwoods: I wasn't sure how we want to handle implementing constant time stuff with the protocol, so these methods are only implemented over `BytesContainer` so far. Because of this the inputs need to be `BytesContainer`s at present.

blochberger:
I like the idea of using the `BytesRepresentable` protocol, lets say for any plaintext. This simplifies the API quite a lot.

I am not so sure about `BytesContainer`, though.
If the intention is just to have something that implements the `BytesRepresentable` protocol I would avoid introducing it and make the types currently wrapped by `BytesContainer` conforming the `BytesRepresentable` protocol.
If the intention is to have a class/structure where the values are confidential, e.g., they require constant time checks, then maybe it is better to make that explicit. Like an abstract class `AbstractSecret` or `SecretBytes` or something. Comparison operators could be implemented in constant time, hex/Base64 en-/decoding as well. And upon destruction of an object the memory could be zeroed. The cryptographic keys could then inherit from this class.

Issue #152, opened 2018-05-28
Linker error in release builds


alexprogrammr:
When using the KeyPairGenerator from Sodium framework that is dynamically linked with main application, the linker dies in release builds with error that it cannot find the protocol witness table for `Sodium.Box : Sodium.KeyPairGenerator` in Sodium.

The main reason for this error is conforming an internal protocol to public class that change visibility of required properties from internal to public. For example:

```
protocol NonceGenerator {
    var NonceBytes: Int { get }
    associatedtype Nonce where Nonce == Bytes
}

public class SecretBox {
    // implementation details
}

extension SecretBox: NonceGenerator {
    public var NonceBytes: Int { return Int(crypto_secretbox_noncebytes()) }
    public typealias Nonce = Bytes
}
```

I have already find related issues with swift compiler that are marked as resolved with appropriate [pull request](https://github.com/apple/swift/pull/15735) but this fix isn't included into latest Swift 4.1 version.

Changing protocol's visibility modifiers to public in whole project will fix this error.

aidantwoods:
Interesting find – going by the dates on some of the linked bugs (around 2016) it appears this isn't limited to Swift 4.1 either.

PR #153, opened 2018-05-28
Make KeyPairGenerator and KeyPairProtocol and related public

Westacular:
This is a quick stopgap to fix https://github.com/jedisct1/swift-sodium/issues/152 where dynamic linking of release builds fail because of Swift compiler issues around a public class conforming to an internal protocol.

### DIFF
--- a/Sodium/Box.swift
+++ b/Sodium/Box.swift
@@ -357,12 +357,12 @@ extension Box: KeyPairGenerator {
     public var PublicKeyBytes: Int { return Int(crypto_box_publickeybytes()) }
     public var SecretKeyBytes: Int { return Int(crypto_box_secretkeybytes()) }
 
-    static let newKeypair: (
+    public static let newKeypair: (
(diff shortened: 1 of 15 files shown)


Sodium/Box.swift
```diff
@@ -373,6 +373,11 @@ extension Box: KeyPairGenerator {
         public typealias SecretKey = Box.SecretKey
         public let publicKey: PublicKey
         public let secretKey: SecretKey
+
+        public init(publicKey: PublicKey, secretKey: SecretKey) {
```
aidantwoods: Isn't this initialiser identical to the [compiler provided one](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Initialization.html#//apple_ref/doc/uid/TP40014097-CH18-ID214)?

tom-wolters:
Is there an ETA when this PR will be merged into master? Would like to work with Xcode 10. Thank you.

Issue #155, opened 2018-06-05
Current version (0.6) fails to compile in Swift 4.2 / Xcode 10


Westacular:
The exclusive access warnings of Swift 4.1 are now errors in Swift 4.2 (https://forums.swift.org/t/upgrading-exclusive-access-warning-to-be-an-error-in-swift-4-2/12704) but it's trivial to hot fix them in 0.6

If the next major version isn't going to be released soon, I'd suggest making a 0.6.1 release with the fixes. I'll make a PR momentarily with the changes.

aidantwoods:
I think these should have been fixed in https://github.com/jedisct1/swift-sodium/pull/137?

Westacular:
Ah, yeah, looks like it's been fixed on master, but my PR uses 0.6 as its base, with the idea that a 0.6.1 release can be made in the meantime for anyone who wants to use both Xcode 10 _and_ a release version of swift-sodium. (But I guess if the next release of swift-sodium is happening soon, then yeah, this is moot.)

taher-mosbah:
@jedisct1 any updates on this ?

lmenus:
+1

brandevnull:
+1

ckempke:
+100  XCode 10 is now the current build of Xcode, so this is failing in production.

mattk90:
If you import the library in your Podfile like this it should compile and work: `pod 'Sodium', :git => 'https://github.com/jedisct1/swift-sodium'`

joejune:
I can confirm that these changes fix the problems, but it would be great if there could be another release that includes these changes.  It would make it possible to have another podspec depend on swift-sodium.  @jedisct1: Would it be possible to create another release?  Here's a PR to help with the podspec file:  #167

Issue #132, opened 2018-03-27
SPM dependency


avi-kik:
It seems that SPM requires a 3-part version tag in order to include a package as a dependency.  I was able to import swift-sodium by adding a `0.6.0` tag in my local copy of the repo, and point my `Package.swift` to that location.  However, this is obviously suboptimal.  Could you add the same tag to this repo and either use such tags exclusively, or in conjunction with, the current two-part tags?

Thanks.

tiwoc:
SPM uses [Semantic Versioning](https://semver.org/), so it is necessary for tags to have the MAJOR.MINOR.PATCH format. It would be great if the following releases could be tagged in this format.

PR #156, opened 2018-06-05
Fix Swift 4.2 exclusive memory access errors

Westacular:
Fix for #155 

This is intended as a hot fix against tag 0.6 but since there's no release branch I can't point the PR at that directly.

Changes are simply to capture the `.count` computed property values into local variables before each `withUnsafeMutableBytes` call to avoid exclusive access errors, as described at https://forums.swift.org/t/upgrading-exclusive-access-warning-to-be-an-error-in-swift-4-2/12704

### DIFF
--- a/Sodium/Box.swift
+++ b/Sodium/Box.swift
@@ -80,8 +80,9 @@ public class Box {
      */
     public func nonce() -> Nonce {
         var nonce = Data(count: NonceBytes)
+        let nonceCount = nonce.count
         nonce.withUnsafeMutableBytes { noncePtr in
-            randombytes_buf(noncePtr, nonce.count)
+            randombytes_buf(noncePtr, nonceCount)
         }
         return nonce
     }
--- a/Sodium/GenericHash.swift
+++ b/Sodium/GenericHash.swift
@@ -50,14 +50,18 @@ public class GenericHash {
         var output = Data(count: outputLength)
         var result: Int32 = -1
 
+        let outputCount = output.count
+        let messageCount = message.count
+
         if let key = key {
+            let keyCount = key.count
             result = output.withUnsafeMutableBytes { outputPtr in
                 message.withUnsafeBytes { messagePtr in
                     key.withUnsafeBytes { keyPtr in
                         crypto_generichash(
-                            outputPtr, output.count,
-                            messagePtr, CUnsignedLongLong(message.count),
-                            keyPtr, key.count)
+                            outputPtr, outputCount,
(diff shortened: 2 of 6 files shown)


aidantwoods:
@jedisct1 I believe the intent here was to commit directly after the 0.6 tag for the purpose of essentially back-porting https://github.com/jedisct1/swift-sodium/pull/137 into a simple patch after the last release. This could then be released as 0.6.1 to push this particular fix out ahead of whenever the next release of `master` happens.

(i.e. if you create a new branch checking out at the 0.6 tag then this PR could instead target that new branch instead of master)

lmenus:
+1

taher-mosbah:
@jedisct1 is it possible to merge this to a branch 0.6.1 like @aidantwoods explained ?
If you are not available I'll be happy to do it if you give me the access needed.

jedisct1:
Why backport it? Would a new release from the current master work?

aidantwoods:
I think the idea was that a backport would be more timely if a new release wasn't happening soon. I would guess that a new release from master would make sense to be 0.7 (and not a patch increment) due to the changes in #144?

jedisct1:
@aidantwoods Yes, #144 introduces a small breaking change, but since it will happen, let it happen now instead of keeping deferring it :)

Do you see anything in the current `master` that would need to be addressed before we can tag a new release?

aidantwoods:
> Do you see anything in the current `master` that would need to be addressed before we can tag a new release?

* Is #150 a desirable change? If so maybe best to deal with it now, since it'll change the common-currency type for swift < 4.1 compatibility—i.e. it touches the same thing we'll already be changing in the next version in another breaking way. Perhaps this kind of abstraction isn't desirable due to timing considerations when converting from byte representable types in a generalised context though?

* (#162) The [`StateStream`](https://github.com/jedisct1/swift-sodium/blob/d918fdbfe95879b1a7c78a9607bc1a03a45fcf22/Sodium/StateStream.swift) protocol (added in #140) is probably now redundant due to the work on removing pointers in #148 (which removed the need for the shared memory freeing methods).

* (#163) Xcode recommends some updates to the project settings

aidantwoods:
Just as a reminder (since it's a little odd), for SPM to function I believe that it requires the next version to be tagged as `0.7.0` instead of `0.7` (#132).

blochberger:
I think this is obsolete mostly due to the changes made by @aidantwoods. I vote for closing this.

PR #158, opened 2018-06-08
Expose features needed to implement the Scuttlebutt Protocol

marching-cube:
In order to implement the Scuttlebutt Protocol, a few more functions need to be exposed from the C Sodium library. Specifically:
- `crypto_scalarmult` is needed for the EC Diffie-Hellman computation,
- `crypto_sign_ed25519_pk_to_curve25519`/ `crypto_sign_ed25519_sk_to_curve25519` are needed so we can use the same keypair for signing and encryption,
- `crypto_hash_sha256` is used instead of the generic BLAKE2b.

Sodium's documentation puts them all in the Advanced section, that's why I am implementing a single Advanced class. Long term, it may be better to split it into separate categories?

Additionally, the current implementation of the SecretBox does not allow user to specify the nonce parameter manually, rather defaults to a random value. However, the protocol requires a fixed nonce value (all zeroes), so I have extended the SecretBox class with an additional `seal` method overload.

Reference:
https://ssbc.github.io/scuttlebutt-protocol-guide/

### DIFF
--- a/Sodium.xcodeproj/project.pbxproj
+++ b/Sodium.xcodeproj/project.pbxproj
@@ -54,6 +54,9 @@
 		90C75EEC1AE3583E00F1E749 /* Sodium.swift in Sources */ = {isa = PBXBuildFile; fileRef = 095D80541A4ECCD7000B83F9 /* Sodium.swift */; };
 		90C75EED1AE3583E00F1E749 /* Utils.swift in Sources */ = {isa = PBXBuildFile; fileRef = 095D805C1A4F4F72000B83F9 /* Utils.swift */; };
 		A0918C781E92489100C1DC33 /* Auth.swift in Sources */ = {isa = PBXBuildFile; fileRef = A0918C771E92489100C1DC33 /* Auth.swift */; };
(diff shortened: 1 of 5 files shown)


jedisct1:
Adding these would be very sad.

swift-sodium tries to retain some sanity by exposing only high-level APIs in order to prevent misuse.

Adding the ability to specify a nonce, use scalarmult without proper hashing, expose a hashing function vulnerable to length extension, completely defeats this.

indutny:
Would exposing `crypto_sign_ed25519_pk_to_curve25519` and `crypto_sign_ed25519_sk_to_curve25519` make the library less high-level? There is certainly a use case for this in my application as well as in secure scuttlebutt.

jedisct1:
@indutny  Conversions between `sodium.sign.KeyPair`, `sodium.box.KeyPair` and `sodium.secretbox.KeyPair` can be added.

None of these should assume specific curves, so they can all fail.

indutny:
I see. Thank you for taking time to explain!

PR #161, opened 2018-08-03
Add functions for converting from Sign.KeyPair to Box.KeyPair

### LINKED ISSUE #160: How to use c function crypto_sign_ed25519_sk_to_curve25519
Hello,

I'm wondering how to call `crypto_sign_ed25519_sk_to_curve25519` and `crypto_sign_ed25519_pk_to_curve25519` from within Swift.  I might be blind but I couldn't see it exposed in any of the swift modules.

Thanks in advance.

MikeEmery:
Fixes #160

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,14 @@ if let unsignedMessage = sodium.sign.open(signedMessage: signedMessage, publicKe
 }
 ```
 
+### Key Conversion
+```swift
+let sodium = Sodium()
+let signingKeyPair = sodium.sign.keyPair()!
+let encryptionKeyPair = sodium.sign.convertEd25519KeyPairToCurve25519(signingKeyPair)!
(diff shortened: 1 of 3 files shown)


jedisct1:
If you can avoid this, please do.

This conversion only works because of the birational equivalence between the curves used by the low-level functions these interfaces currently use. This is unlikely to be the case with other constructions.

There is also no proof that doing so is safe.

This requires a low-level function, and the Swift bindings intentionally try to keep things sane and only support high-level functions that are unlikely to be deprecated, removed or changed later. And now that libsodium directly supports DH over Ed25519, this conversion function is likely to go away.

Both types of keys can be derived from a seed (`keypairFromSeed`). If you need to do both with a single secret key, use these functions to create key pairs instead.

PR #167, opened 2018-10-02
Update podspec for 0.7 release.

joejune:
Once this is merged, the `0.7` tag will need to be added, and the updated podspec will need to be submitted to the Cocoapods spec repo.

### DIFF
--- a/Sodium.podspec
+++ b/Sodium.podspec
@@ -7,7 +7,7 @@ s.homepage = 'https://github.com/jedisct1/swift-sodium'
 s.social_media_url = 'https://twitter.com/jedisct1'
 s.authors = { 'Frank Denis' => '' }
 s.source = { :git => 'https://github.com/jedisct1/swift-sodium.git',
-             :tag => '0.6' }
+             :tag => '0.7' }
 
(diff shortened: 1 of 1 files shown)


jedisct1:
You forgot the Carthage package :)

joejune:
I'm not sure what you mean.  When specifying a Carthage dependency, you just indicate a repo and a version(commit-like).  So if someone wanted to update their Carthage dependency, they'd just change the tag to be `0.7` after you've created the new tag.  

If I'm overlooking something else I'll be happy to fix it, but you'll need to at least point me to a specific file.

Thanks!

jedisct1:
The `Sodium.framework.zip` file still has to be built.

Anyway, a pull request is not required to tag a new version.

What's more required is make sure it works. Current blockers are:

- issue with bitcode (to be confirmed)
- possibly support for WatchOS (people asked for this, but nobody knows how to build the library for it)

joejune:
Are either of those two blockers you mentioned regressions from the 0.6 release?  If they are not, I don't think they should be considered blockers.

PR #173, opened 2018-11-25
Add parameter to specify nonce

aidantwoods:
It seems that the current Swift API doesn't permit this currently. I wonder if this could be included in a way such that is is slightly more hidden away from the main AEAD API?
This might be too much of a foot-gun to justify including, so if there are concerns I'm happy for the result of this PR to be that it is closed and not merged.

### DIFF
--- a/Sodium/Aead.swift
+++ b/Sodium/Aead.swift
@@ -37,16 +37,18 @@ extension Aead.XChaCha20Poly1305Ietf {
 
      - Parameter message: The message to encrypt.
      - Parameter secretKey: The shared secret key.
+     - Parameter nonce: The nonce, NEVER reuse with the same secretKey.
      - Parameter additionalData: A typical use for these data is to authenticate version numbers, timestamps or monotonically increasing counters
 
-     - Returns: The authenticated ciphertext and encryption nonce.
+     - Returns: The authenticated ciphertext.
      */
-    public func encrypt(message: Bytes, secretKey: Key, additionalData: Bytes? = nil) -> (authenticatedCipherText: Bytes, nonce: Nonce)? {
-        guard secretKey.count == KeyBytes else { return nil }
+    public func encrypt(message: Bytes, secretKey: Key, nonce: Nonce, additionalData: Bytes? = nil) -> Bytes? {
+        guard secretKey.count == KeyBytes,
+            nonce.count == NonceBytes
+        else { return nil }
 
         var authenticatedCipherText = Bytes(count: message.count + ABytes)
         var authenticatedCipherTextLen: UInt64 = 0
-        let nonce = self.nonce()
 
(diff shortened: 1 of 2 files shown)


davidperrenoud:
To solve this problem, would it be possible to expose the low-level libsodium API in Swift-Sodium?

Currently, [Swift-Paseto](https://github.com/aidantwoods/swift-paseto) depends on [Clibsodium](https://github.com/tiwoc/Clibsodium) which results in unnecessary complications, as the libsodium headers are included twice and this could result in a version mismatch.

I cannot give an opinion as for requiring a specific nonce in an AEAD construction, but it is clearly required on page 13 of the [PASETO RFC](https://paseto.io/rfc/):
```
   3.  Calculate BLAKE2b of the message "m" with the output of step 2 as
       the key, with an output length of 24.  This will be our nonce,
       "n".

       *  This step is to ensure that an RNG failure does not result in
          a nonce-misuse condition that breaks the security of our
          stream cipher.

   4.  Pack "h", "n", and "f" together (in that order) using PAE (see
       Section 2.2).  We'll call this "preAuth".

   5.  Encrypt the message using XChaCha20-Poly1305, using an AEAD
       interface such as the one provided in libsodium.  (See below for
       pseudocode.)
```

jedisct1:
The nonce has to be unique for a given key. Or, at least, the same nonce shouldn't be used with the same key but different messages.

The PASETO recommendation you mention is just a mitigation for a failure of step 2, i.e. a random number generator that would return the same value twice.

This will not happen with the libsodium RNG, unless live VM images are cloned, and the hypervisor doesn't provide a random device, and the operating system doesn't reseed the RNG. But on such an environment, your secret keys are at risk already.

So, using swift-sodium, the nonces are indistinguishable from random already. You don't have to make them "more random".

davidperrenoud:
It seems that @paragonie-security [already updated the PASETO draft](https://github.com/paragonie/paseto/blob/master/docs/RFC/draft-paragon-paseto-rfc-01.txt#L733) to make nonce-misuse resistance optional if the library already takes care of it:
```
   3.  Optionally, calculate BLAKE2b of the message "m" with the output
       of step 2 as the key, with an output length of 24.  This will be
       our nonce, "n".

       *  This step is to ensure that an RNG failure does not result in
          a nonce-misuse condition that breaks the security of our
          stream cipher.

       *  If this step is omitted, the output of step 2 is "n" instead.
```

I am a bit confused as to why `draft-01` is not part of the official website yet.

Note that dropping the nonce parameter from the libsodium's public API would have an impact on most implementations of PASETO, as they all depend on libsodium in one form or another.

Related issues:
https://github.com/paragonie/paseto/pull/23
https://github.com/paragonie/paseto/pull/76

Issue #177, opened 2019-01-28
Does not build with Xcode 10.2 beta 1, "Use of undeclared type 'crypto_generichash_state'"


Westacular:
When building with Xcode 10.2 beta 1, GenericHash.swift hits errors related to `Use of undeclared type 'crypto_generichash_state'`

The underlying issue appears to be with
```c
typedef struct CRYPTO_ALIGN(64) crypto_generichash_blake2b_state {
```

Specifically: the Swift compiler now [limits maximum type alignment size at 16](https://github.com/apple/swift/commit/c1f99b5fdb717ca6135515ac94f32a5ffdd17e34), and any C type that specifies a greater alignment requirement gets imported to Swift as `OpaquePointer`.

lmenus:
@Westacular Have you been able to get this working?

davidperrenoud:
I have been able to make it compile with the following changes in `GenericHash.swift`:
```swift
        // ...
        private var opaqueState: OpaquePointer
        public var outputLength: Int = 0

        init?(key: Bytes?, outputLength: Int) {
            state = UnsafeMutableRawBufferPointer.allocate(byteCount: crypto_generichash_statebytes(), alignment: 64)
            guard state.baseAddress != nil else { return nil }
            opaqueState = OpaquePointer(state.baseAddress!)
        // ...
```

All the tests pass but could someone check if this is correct?

marwan87:
@davidperrenoud, im having the same issue, if I use your code the compiler still doesn't the line, "private var state: crypto_generichash_state" Thanks for any help.

keefertaylor:
I have a similar error building with the non-beta Swift 5 / XCode 10.2 toolchain:

```
swift-sodium/Sodium/GenericHash.swift:17:55: error: use of undeclared type 'crypto_generichash_state'
        private var opaqueState: UnsafeMutablePointer<crypto_generichash_state>
```

alexisbronchart:
@marwan87 are you using master?

davidperrenoud:
@jedisct1 Could you reopen this issue?

marwan87:
@alexisbronchart  I ended up getting it work by using 
```pod "Sodium", :git => 'https://github.com/Westacular/swift-sodium.git', :branch => 'fix_xcode10.2'```
that looks like its a good fix 
its highlighted in https://github.com/jedisct1/swift-sodium/pull/178
[...]

PR #178, opened 2019-01-28
Use sodium_malloc(crypto_generichash_statebytes()) to alloc hash state struct

Westacular:
Addresses https://github.com/jedisct1/swift-sodium/issues/177

As of Xcode 10.2 beta 1 (i.e. the Swift 5 compiler, even in Swift 4 mode), Swift no longer bridges C types which have alignment sizes > 16 bytes, so instead of trying to alloc the `crypto_generichash_state` struct in Swift, malloc it using the recommended C functions and treat it as an `UnsafeMutableRawPointer`.

Wrapping it as a `OpaquePointer` is necessary for interacting with the other C APIs, but we need to also keep the original `UnsafeMutableRawPointer` in order to free the memory in deinit.

### DIFF
--- a/Sodium/GenericHash.swift
+++ b/Sodium/GenericHash.swift
@@ -13,26 +13,45 @@ public struct GenericHash {
 
 extension GenericHash {
     public class Stream {
-        private var opaqueState: OpaquePointer?
+    #if compiler(>=5)
+        private var opaqueState: OpaquePointer
         private var state: UnsafeMutableRawPointer!
+    #else
+        private var state: crypto_generichash_state
+    #endif
+
         public var outputLength: Int = 0
 
         init?(key: Bytes?, outputLength: Int) {
+
+        #if compiler(>=5)
             state = sodium_malloc(crypto_generichash_statebytes())
-            opaqueState = OpaquePointer(state)
+            guard let opaqueState = OpaquePointer(state) else { return nil }
+            self.opaqueState = opaqueState
 
             guard .SUCCESS == crypto_generichash_init(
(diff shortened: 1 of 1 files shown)


jedisct1:
What happens exactly with types having a > 16 bytes alignment? If they end up being 16 bytes aligned, this is fine.

Westacular:
Any struct where the C attributes specify an alignment requirement > 16 bytes gets treated as something that can't be represented in Swift — the struct itself is not imported, and all parameters that use it as a type instead show up as type `OpaquePointer?` in Swift. Exactly 16 is fine, however.

I've updated my PR such that it's compatible with both Xcode 10.0/1 and 10.2 (yay `#if compiler(>=5)`), but if crypto_generichash_blake2b.h can instead be modified to specify
```c
typedef struct CRYPTO_ALIGN(16) crypto_generichash_blake2b_state
```
that would be a nicer fix. (I didn't know enough about the underlying C implementation to guess at why it specified alignment of 64, or if it would be safe to change.)

marwan87:
Is this fix going to be merged into the master branch

rcamparoML:
Why this PR wasn't merged ? :(

jedisct1:
This is redundant with 7de099998e5184c941c3723957157217b41829bd, and alignment is not required any more.

jedisct1:
Backporting the code to obsolete Xcode versions should be pretty simple. Not sure if the precompiled binaries will be compatible, though. Probably, but this is completely untested and unsupported, so you're a bit on your own here.

PR #181, opened 2019-03-28
Ensure examples compile on Xcode 10.2

Rypac:
Add `#availability` checks to ensure examples are Xcode 10.2 compatible.

### DIFF
--- a/Examples/CommonCode/Helpers.swift
+++ b/Examples/CommonCode/Helpers.swift
@@ -8,7 +8,11 @@ extension String {
 
 extension Dictionary {
     func toData() -> Data? {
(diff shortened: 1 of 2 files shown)


jedisct1:
Thanks!

Don't we need similar checks for macOS and watchOS?

Rypac:
> Don't we need similar checks for macOS and watchOS?

Yes, for macOS we do but the watchOS example targets 5.0 which supports the new method.

I've added a guard for the macOS example and bumped the deployment target to 10.11. Seems the run fine for all platforms now.

jedisct1:
Thank you!

PR #183, opened 2019-05-03
Add HMAC-SHA-256/512/512-256 and support for multi-part message HMACs

FabioTacke:
This implements the libsodium HMAC API which adds support for HMAC-SHA-512 and HMAC-SHA-256 (and also HMAC-SHA-512-256, but that has already been availably via `sodium.auth.tag()`).

```swift
let sodium = Sodium()
let message = "My Test Message".bytes
let key = sodium.hmac.key(alg: .sha256)!
let hmac = sodium.hmac.authenticate(message: message, key: key, alg: .sha256)!
let hmacIsValid = sodium.hmac.verify(hmac: hmac, message: message, key: key, alg: .sha256)
```

Furthermore it enables the HMAC calculation of multi-part messages (e.g. for streaming data).

```swift
let sodium = Sodium()
let key = sodium.hmac.key(alg: .sha256)!
let multiPart = sodium.hmac.initMultiPartHMAC(key: key, alg: .sha256)!

let inputPart1 = "This is a".bytes
multiPart.update(message: inputPart1)

let inputPart2 = "multi-part message".bytes
multiPart.update(message: inputPart2)

let hmac = multiPart.final()!

let completeMessage = "This is a multi-part message".bytes
let hmacIsValid = sodium.hmac.verify(hmac: hmac, message: completeMessage, key: key, alg: .sha256)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,8 @@ if sodium.pwHash.strNeedsRehash(hash: hashedStr,
 
 ## Authentication tags
 
+### Single-part messages
+
 The `sodium.auth.tag()` function computes an authentication tag (HMAC) using a message and a key. Parties knowing the key can then verify the authenticity of the message using the same parameters and the `sodium.auth.verify()` function.
 
 Authentication tags are not signatures: the same key is used both for computing and verifying a tag. Therefore, verifiers can also compute tags for arbitrary messages.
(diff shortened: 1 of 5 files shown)


jedisct1:
Hi!

Hashing, including keyed hashing, is already available as `sodium.genericHash.*`, and this includes support for multi-part messages.

HMAC would be redundant. Swift-sodium tries to avoid low-level APIs that may eventually be removed from the core library.

PR #184, opened 2019-05-06
fix enum to lower case

chinsyo:


Sodium/SecretStream.swift
```diff
@@ -16,10 +16,10 @@ extension SecretStream {
 
 extension SecretStream.XChaCha20Poly1305 {
     public enum Tag: UInt8 {
-        case MESSAGE = 0x00
-        case PUSH = 0x01
-        case REKEY = 0x02
-        case FINAL = 0x03
+        case message = 0x00
+        case push = 0x01
+        case pekey = 0x02
```
blochberger: Typo: `pekey` instead of `rekey`

jedisct1:
But... why?

It breaks all the tests and even the examples from the documentation.

chinsyo:
> But... why?
> 
> It breaks all the tests and even the examples from the documentation.

emmm... Cuz swift standard library do.

Sodium/PWHash.swift
```diff
@@ -109,16 +109,16 @@ extension PWHash {
      - Parameter salt: Unpredicatable salt data.  Must have a fixed length of `SaltBytes`.
      - Parameter opsLimit: Represents a maximum amount of computations to perform. Raising this number will make the function require more CPU cycles to compute a key.
      - Parameter memLimit: The maximum amount of RAM that the function will use, in bytes.
-     - Parameter alg: The algorithm identifier (`.Default`, `.Argon2I13`, `.Argon2ID13`).
+     - Parameter alg: The algorithm identifier (`.eefault`, `.argon2I13`, `.argon2ID13`).
```
jedisct1: `eefault`?

Sodium/SecretStream.swift
chinsyo: Addressed.

Sodium/PWHash.swift
chinsyo: > `eefault`?

Sorry for that, addressed.

jedisct1:
But the current names are consistent with the underlying C library.

Following Swift's convention is good, but this change is going to break all existing apps without any functional improvement.

I'll let @blochberger judge if it's worth it. But if such a change is made, it has to be made everywhere, including tests, examples and documentation.

chinsyo:
> But the current names are consistent with the underlying C library.
> 
> Following Swift's convention is good, but this change is going to break all existing apps without any functional improvement.
> 
> I'll let @blochberger judge if it's worth it. But if such a change is made, it has to be made everywhere, including tests, examples and documentation.

Thanks for your reply.

blochberger:
> But the current names are consistent with the underlying C library.

This is an implementation detail and in my case only relevant if references from the [Sodium documentation](https://download.libsodium.org/doc/) would cause confusion – which I do not think is the case.

> Following Swift's convention is good, but this change is going to break all existing apps without any functional improvement.

I think that this is a good point. However, some parts/ideas of the pull request might be worth giving it a bit more thought:
- The `SUCCESS`/`FAILURE` cases could be replaced by the `Result` type introduced in Swift 5 ([SE-0235](https://github.com/apple/swift-evolution/blob/master/proposals/0235-add-result.md)).
- `default` is a reserved keyword requiring special quoting – as your change shows. I think this is confusing and should be avoided. Maybe the `Default` case can be dropped in the enum altogether or renamed to `defaultAlgorithm`. Dropping the case requires more thought. Simply choosing the default algorithm for any unknown integer passed to the `switch` statement would be a bad choice. Calling the underlaying C-function in the function signature (this is what I do in [Tafelsalz](https://github.com/blochberger/Tafelsalz/blob/903bba7e6536512cf3e0f34efce1283b7eae5ff1/Tafelsalz/Sodium.swift#L650)) might also not be the best choice.

> I'll let @blochberger judge if it's worth it. But if such a change is made, it has to be made everywhere, including tests, examples and documentation.

I vote for closing the pull request for now. I could open issues for using the new `Result` type and for discussing how the `Default` algorithm could be handled more "swifty" – which I do not think is straight forward. What do you think @jedisct1?

jedisct1:
If we are going to make breaking changes, using the `Result` type definitely looks like a good change to make.

Using quotes to work around existing keywords is painful. Renaming `Default` to `defaultAlgorithm` would work.

blochberger:
Looking at `ExitCode` again, I see that it is only used internally, i.e., not communicated to the user. Regardless of which error occurred, `nil` is returned for all fallible functions. Hence, I think that using the `Result` type is not needed in that case.

chinsyo:
> Looking at `ExitCode` again, I see that it is only used internally, i.e., not communicated to the user. Regardless of which error occurred, `nil` is returned for all fallible functions. Hence, I think that using the `Result` type is not needed in that case.

Agree

GoranLilja:
What's the status of this pull request? I got the impression that everyone is agreeing, yet the PR isn't merged. Friendly ping @blochberger who requested changes. 🙂

Sodium/PWHash.swift
```diff
@@ -15,18 +15,18 @@ public struct PWHash {
 
 extension PWHash {
     public enum Alg {
-        case Default
-        case Argon2I13
-        case Argon2ID13
+        case `default`
```
blochberger: Avoid ticks and name it `defaultAlgorithm` as @jedisct1 [suggested](https://github.com/jedisct1/swift-sodium/pull/184#issuecomment-489597858).

Sodium/PWHash.swift
```diff
@@ -15,18 +15,18 @@ public struct PWHash {
...
     }
 }
 
 extension PWHash.Alg {
     var id: Int32 {
         switch self {
-        case .Default:    return crypto_pwhash_alg_default()
-        case .Argon2I13:  return crypto_pwhash_alg_argon2i13()
-        case .Argon2ID13: return crypto_pwhash_alg_argon2id13()
+        case .`default`:    return crypto_pwhash_alg_default()
```
blochberger: Avoid ticks and name it `defaultAlgorithm` as @jedisct1 [suggested](https://github.com/jedisct1/swift-sodium/pull/184#issuecomment-489597858).

PR #188, opened 2019-08-04
Add nonce-accepting overload of `secretBox.seal`

indutny:
As good as the current API is at some times it is necessary to have
reproducible results of the encryption. This commit introduces:

```
sodium.secureBox.seal(message: [...], secretKey: [...], nonce: [...])
```

That takes the usual `message` and `secretKey`, and an additional
user-provided `nonce` to generate the result that is completely defined
by those inputs and does not rely on any entropy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,8 @@ This API encrypts a message. The decryption process check that they haven't been
 
 Messages encrypted that way are independent: if multiple messages are sent that way, the recipient cannot detect if some messages have been duplicated, deleted or reordered without including additional data to each message.
 
+Optionally, `SecretBox` provides the ability to utilize a user-defined nonce via `seal(message: secretKey: nonce:)`.
+
 ## Public-key Cryptography
 
 With public-key cryptography, each peer has two keys: a secret key, that has to remain secret, and a public key that anyone can use to send an encrypted message to that peer. That public key can be only be used to encrypt a message. The related secret is required to decrypt it.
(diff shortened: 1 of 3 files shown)


jedisct1:
Hi!

You can use the deterministic number generator in order to get a reproducible key stream.

In swift-sodium, the nonce is not something applications should have to care about.

indutny:
Hello!

I appreciate promptness of your review. However, I'd like to ask you to take another look at this PR for following reasons:

* `sodium.box.seal` already supports `nonce` argument: https://github.com/jedisct1/swift-sodium/blob/419a1c0fbf307bcf75d177d042649dca7fdb3fbc/Sodium/Box.swift#L41
* My use case is not related to writing tests, but is rather related to distributed systems. I'd like messages in my system to have consistent hashes, even if I re-encrypt them.

Thank you!

jedisct1:
Oh, that's atrocious :(

Can you update the documentation in that PR as well?

indutny:
@jedisct1 sure thing. Updated commit message. Thank you!

jedisct1:
Err... the [documentation](https://github.com/jedisct1/swift-sodium/blob/master/README.md)

indutny:
Aaah... My bad. Force pushed the commit again. Let me know if this is what you meant.

(I literally copied it from `Box` section of README.md, in case if it matters)

indutny:
🎉  Thank you! 🎉

PR #194, opened 2019-12-20
Clibsodium defined as an explicit module and Clibsodium import statem…

### LINKED ISSUE #190: Package Manager - could not build Clibsodium
Hey!

First off, thanks for this project :) 

Now, I wanted to add this library as a dependency to my package, but it fails with two errors:

In `Aead.swift` it can not `import Clibsodium` with this error: "Could not build Objective-C module 'Clibsodium'"

And in `Clibsodium`, I get "'sodium.h' file not found" in `shimmer.h`.

Running `brew install clibsodium` helped, but then when I try to run example app with my package on the iOS simulator, I get this error: 
"dyld: Library not loaded: /usr/local/opt/libsodium/lib/libsodium.23.dylib
  Referenced from: Path-To-Example
  Reason: no suitable image found.  Did find:
	/usr/local/opt/libsodium/lib/libsodium.23.dylib: mach-o, but not built for iOS simulator
"

Is there something I am doing wrong? Are there any additional steps needed to add `swift-sodium` to a package that supports iOS?

FedericoHenze:
Fix #190 

Clibsodium defined as an explicit module and Clibsodium import statements changes to access the clibsodium via the Sodium framework.

Inspired on:
- https://badootech.badoo.com/bundling-c-library-in-swift-framework-3d9dae950774

> In order to import C library into Swift, it first need to be be declared as a module. Declaration takes the form of a module map — a file describing headers to import and static libraries to link with. The resulting module may be imported into Swift (natively) and Objective-C.
> 
> This way of embedding C library into a framework will work for you in most cases (learn more about this approach here and there). It is totally legit as long as you’re developing some internal framework or simply modularising your application into separate parts. However, it doesn’t work well when you need to ship your library to external users via Carthage, Cocoapods or as a binary. The main reason being that the resulting framework is not portable between computers. When undergoing compilation, your resulting framework is bound to the current location of headers and libraries from module map on your computer. You can use it right away in your project but should you try to send it to somebody else they will not be able to link it to a project because files referenced by module map are no longer accessible.


marcelofabri:
Curious why this got reverted @jedisct1

jedisct1:
Because it didn't compile: https://travis-ci.com/jedisct1/swift-sodium/builds/141938692

PR #197, opened 2020-01-20
Validate CocoaPods and project builds in CI

marcelofabri:
- Removing `.swift-version` was required for `pod lib lint` to pass as we use the new attribute in the Podspec.
- `SUPPORTS_MACCATALYST=NO` was required for `xcodebuild test` succeed as the `.a` library is not compiled for Catalyst.


jedisct1:
Thanks!

PR #198, opened 2020-01-20
Clibsodium defined as an explicit module and Clibsodium import

marcelofabri:
This reverts the revert of #194 and add an extra commit to make it work with Swift Package Manager by using conditional imports


jedisct1:
The `#if` statements added everywhere for `Clibsodium` are not great. Is it really the only way?

The module map is actually a bit more concerning. It's very likely that it will not be updated when files are added/removed/renamed.
Could it be automatically created using a script? Maybe in CI, its content should also be compared with the files present in the directory.

jedisct1:
Related: now that SPM has been around for quite some time, do we still have to support Cocoapods?

marcelofabri:
> Related: now that SPM has been around for quite some time, do we still have to support Cocoapods?

I'd recommend keeping it for now, as SPM has been usable on iOS only for a few months and still doesn't support assets or binary frameworks, so a lot of projects still can't use it.


> The `#if` statements added everywhere for `Clibsodium` are not great. Is it really the only way?

Not sure if they're the only way, but it was an easy way to get this to the finish line as it was so close. This is very important for us because it allows to share Carthage cached frameworks between developers in the team. Right now, we have a workaround for swift-sodium to manually build it with Carthage, which ends up taking 3-5 minutes on every CI build (and sometimes on local dev machines).

> 
> The module map is actually a bit more concerning. It's very likely that it will not be updated when files are added/removed/renamed.

CI will fail now that #197 was merged if this happens.

> Could it be automatically created using a script? Maybe in CI, its content should also be compared with the files present in the directory.

It could, but I'm not sure it's worth it right now. I imagine that it's not _that_ often that headers are added/removed in libsodium and we'd get a CI failure if this happens.

ashchan:
Does this still require a libsodium installation on the target macOS platform if a project uses SPM to depend on swift-sodium? Thanks.

marcelofabri:
> Does this still require a libsodium installation on the target macOS platform if a project uses SPM to depend on swift-sodium? Thanks.

I think so, this should have no impact in SPM

ashchan:
@marcelofabri Thanks! I guess I had a misunderstanding that since there're `libsodium-ios.a` and `libsodium-osx.a`, we might be able to link without `brew install libsodium` on macOS 😿.

marcelofabri:
> @marcelofabri Thanks! I guess I had a misunderstanding that since there're `libsodium-ios.a` and `libsodium-osx.a`, we might be able to link without `brew install libsodium` on macOS 😿.

It looks like the issue is that SPM doesn't support binary dependencies (`.a`) so there's no way to avoid that (but I have no context in this particular case, so I might be missing something)

ashchan:
> > @marcelofabri Thanks! I guess I had a misunderstanding that since there're `libsodium-ios.a` and `libsodium-osx.a`, we might be able to link without `brew install libsodium` on macOS 😿.
> 
> It looks like the issue is that SPM doesn't support binary dependencies (`.a`) so there's no way to avoid that (but I have no context in this particular case, so I might be missing something)

That seems so. Hope one day we can do that with SPM.

jedisct1:
I don't understand what `Clibsodium` is and does.

It seems to link against a system-wide installation of `libsodium` installed using homebrew. But that one doesn't necessarily match the version expected by `swift-sodium`, and doesn't include any support for iOS and watchOS.

jedisct1:
With these changes, the library doesn't build any more:

```text
No such module 'Sodium.Clibsodium'
```

jedisct1:
With these changes, the library doesn't build any more:

```text
No such module 'Sodium.Clibsodium'
```

jedisct1:
This is when trying to build any example.

marcelofabri:
> I don't understand what `Clibsodium` is and does.
> 
> It seems to link against a system-wide installation of `libsodium` installed using homebrew. But that one doesn't necessarily match the version expected by `swift-sodium`, and doesn't include any support for iOS and watchOS.

SPM still uses https://github.com/tiwoc/Clibsodium.git and non-SPM uses the `libsodium.a`


> This is when trying to build any example.

Interesting. The framework target definitely builds (we've been using my fork with Carthage), so I wonder what's different in the Example project. I'll try to investigate later.

marcelofabri:
OK, the issue was that the watchOS framework was not properly setup. Fixed it (and rebased against `master`).

marcelofabri:
Is there anything I can do to help getting this merged? 🚀

PR #200, opened 2020-05-06
Correct constructor call

own2pwn:
`(())!` Passing void with unwrapping seems to be outdated


own2pwn:
funny enough I've committed with typo as well and that's been the 200th... fixed in #202

PR #201, opened 2020-05-06
Add hmacsha256 auth functions

justinsinger:
I needed to use `crypto_auth_hmacsha256` in my Swift app and couldn't find a way to do it. This PR doesn't include the tagging methods for  HMAC-SHA-512 or HMAC-SHA512-256 (truncated HMAC-SHA-512), but consumers of this library may want to use those in the future as well. Let me know if I missed a way to use `crypto_auth_hmacsha256` with the existing Swift bindings, or if the purpose of this project it to provide a simplified subset of the functionality of LibSodium.

### DIFF
--- a/Sodium/Auth.swift
+++ b/Sodium/Auth.swift
@@ -47,6 +47,47 @@ extension Auth {
             secretKey
         ).exitCode
     }
+
+    /**
+     Computes an authentication tag for a message using a key
+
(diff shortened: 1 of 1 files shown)


jedisct1:
These bindings are only for the stable, high-level APIs.

`crypto_auth_hmacsha256` is not one of these, sorry.

PR #203, opened 2020-05-17
added: box.seal(message:beforenm:nonce:)

projectvim:


jedisct1:
But... why?

projectvim:
I am porting to swift some library, that needs that. Maybe someone else also will need it.

Issue #213, opened 2020-10-06
Pod fails validation with Xcode 12 toolchain


mokagio:
Hi all 👋 Thank you for the great work building and maintaining this library. We use it in [one of our libraries at Automattic](https://github.com/Automattic/Automattic-Tracks-iOS/blob/5f96537a62a8b7d3eb6635cb4dd8489dd39fc62a/Automattic-Tracks-iOS.podspec#L35).

I'm experiencing an issue with the Xcode 12 toolchain and CocoaPods: the `.podspec` validation fails with this error:

```
❌ ld: in /path/to/CocoaPods-Lint/workspace/Pods/Sodium/Sodium/libsodium/libsodium-ios.a(libsodium_la-aead_xchacha20poly1305.o), 
building for iOS Simulator, 
but linking in object file built for iOS,
for architecture arm64

❌ clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

You can reproduce this on `0.8.0` by running:

```
pod spec lint ./Sodium.podspec
```

I been able to get the validation to pass for version `0.8.0` for iOS and macOS, but not for watchOS. [Here's the branch on my fork](https://github.com/mokagio/swift-sodium/tree/exclude-arm64-architecture) if you want to check it out.

But, `master` has changed the way `libsodium` is imported with the `.xcframework` (which is pretty neat) and unfortunately the pod validation fails there, too 😞 .

I tried with CocoaPods 1.9.3 and also 1.10.0.rc.1. Here's the validation output from the latter:

```
- NOTE  | [iOS] xcodebuild:  warning:
  [CP] Unable to find matching .xcframework slice in 
  'Sodium/Clibsodium.xcframework Clibsodium library
  ios-arm64_x86_64-maccatalyst ios-i386_x86_64-simulator
  ios-arm64_armv7_armv7s' for the current build architectures
  (arm64 i386 x86_64).
- ERROR | xcodebuild:  Sodium/Sodium/Aead.swift:2:8: error: 
  no such module 'Clibsodium'
- NOTE  | [OSX] xcodebuild:  clang: error:
  linker command failed with exit code 1 (use -v to see invocation)
- NOTE  | [watchOS] xcodebuild:  warning:
 [CP] Unable to find matching .xcframework slice in 
 'Sodium/Clibsodium.xcframework Clibsodium library 
 watchos-i386_x86_64-simulator watchos-arm64_32_armv7k'
 for the current build architectures (arm64 i386 x86_64).
```

[...]

PR #214, opened 2020-10-08
Explicitly exclude arm64 architecture in .podspec

mokagio:
Relates to #213.

Xcode 12 changed the way apps are built because of the support for Apple Silicon (ARM chip). We need to explicitly exclude the arm64 architecture.

More info in [this StackOverflow question](https://stackoverflow.com/questions/63607158/xcode-12-building-for-ios-simulator-but-linking-in-object-file-built-for-ios).

Unfortunately **this doesn't yet fix the validation of the spec**. Nevertheless, even though I could be wrong, I think it's a step forward in the right direction.

### DIFF
--- a/Sodium.podspec
+++ b/Sodium.podspec
@@ -19,12 +19,25 @@ s.watchos.deployment_target = '5.0'
 s.source_files = 'Sodium/**/*.{swift,h}'
 s.private_header_files = 'Sodium/libsodium/*.h'
 
+# Xcode 12 changed the way apps are built because of the upcoming support for
+# Apple Silicon. We need explicitly exclude the arm64 architecture.
+# More info at
+# https://stackoverflow.com/questions/63607158/xcode-12-building-for-ios-simulator-but-linking-in-object-file-built-for-ios
 s.pod_target_xcconfig = {
-       'SWIFT_INCLUDE_PATHS' => '$(inherited) "${PODS_XCFRAMEWORKS_BUILD_DIR}/Clibsodium"'
+  'SWIFT_INCLUDE_PATHS' => '$(inherited) "${PODS_XCFRAMEWORKS_BUILD_DIR}/Clibsodium"',
+  'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'arm64',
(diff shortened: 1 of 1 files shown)


jedisct1:
Uploading a release to the Cocoapods repository doesn't work:

```text
 -> Sodium (0.9.0)
    - ERROR | xcodebuild: Returned an unsuccessful exit code. You can use `--verbose` for more information.
    - ERROR | xcodebuild:  Sodium/Sodium/Aead.swift:2:8: error: no such module 'Clibsodium'
```

They probably need to update the code on their end, so let's wait until the final 1.10 version is released.

mokagio:
Hey @jedisct1, thanks for taking a look a this, and for the great work on the Sodium project 👏 👍 💜 .

Yeah, unfortunately this doesn't yet fix the validation of the spec. Nevertheless, even though I could be wrong, I think it's a step forward in the right direction.

I tried this with the Xcode 12 toolchain and CocoaPods 1.10.0.rc.1 and still couldn't get pass the validation.

The error we're getting right now should be

```
ld: library not found for -lClibsodium
```

I don't understand how that's happening since I can see the imports in the library search paths etc in the `.xcconfig` that CocoaPods generates. I must be missing something.

jedisct1:
It's a bug in Cocoapods that was fixed after `rc1`.

mokagio:
Thanks for the info @jedisct1. You are right. I tried with `cocoapods` built from their main branch and it passed the validation 🎉 .

I also tried reverting the change in this PR, and the validation failed, so I believe it'll be appropriate to merge this and release a new version on CocoaPods after testing on 1.10.0 once they publish it.

brow:
[CocoaPods 1.10.0 is released! 🎉](https://github.com/CocoaPods/CocoaPods/releases/tag/1.10.0)

jedisct1:
Yes, I tried about 10 minutes after the release.

Validation now works locally, but the pods can't be uploaded to their servers. They are probably still running an old version.

jedisct1:
Uploaded, but without support for Apple Silicon :(

mokagio:
Thank you for this!

PR #215, opened 2020-10-09
Ensure Package.swift and Sodium.podspec both specify Swift 5.0

danielrbrowne:
This PR:
- Sets `swift-tools-version` in `Package.swift` to 5.0 (to match `swift_version` setting in `Sodium.podspec`).

**N.B:**
- Reducing the `swift-tools-version` setting from 5.3 to 5.0 (to match the podspec file) also allows the library to be integrated using the 'Swift Packages' integration present in Xcode 11.x and above. (Whereas leaving it set at 5.3 meant that Xcode 12.x and above was required).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.0
 import PackageDescription
 
 let package = Package(


danielrbrowne:
@jedisct1 Not sure why Travis failed here - looks to be a spurious failure fetching `homebrew`. Does this seem like a sensible change? My motivation was that the SDK I maintain currently builds using Xcode 11.0 and above, and I'd like to update to swift-sodium 0.9.0 (for SPM fixes) without having to increase our minimum Xcode version to 12.0

jedisct1:
I can certainly tag a new release now if you need this :)

jedisct1:
Still, can you check that everything works as expected on tvOS?

Having an example (like the other ones in the `Examples` directory) for tvOS would also be wonderful.

danielrbrowne:
That would be great - just a bit cleaner than referencing `master` IMHO (even if no additional changes need pushing).

danielrbrowne:
I will open a PR with a new example to confirm it works before you tag the release.

jedisct1:
Awesome, thank you!

PR #216, opened 2020-10-14
Add tvOS support

danielrbrowne:
This PR adds tvOS support, specifically the following are added:

- Adds `Sodium_tvOS` framework build target and scheme.
- Adds `Example tvOS` app build target and scheme
  - This uses the same code as the `Example iOS` app in `ViewController.swift` to test that basic functionality is working as expected when running on tvOS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ build/
 xcuserdata
 .build
 Package.resolved
+.DS_Store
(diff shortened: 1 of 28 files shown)


danielrbrowne:
@jedisct1 Here's the PR I promised. In my testing I get no issues :) If you're happy with it, I'd be grateful for a new tagged release. You might need to update the podspec file and/or `Package.swift` to reflect the tvOS support but I left it from this PR.

PR #219, opened 2020-10-14
Resolve Xcode 12 Package manifest resolution error

danielrbrowne:
This PR is effectively a revert of the PR https://github.com/jedisct1/swift-sodium/pull/215 (which was made in error it turns out):

- Sets `// swift-tools-version:5.3` in `Package.swift` since `.binaryTarget(…)` is only available in Xcode 12.0 and above.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.0
+// swift-tools-version:5.3
 import PackageDescription
 
 let package = Package(


danielrbrowne:
I've confirmed this works integrating with Xcode 12's 'Swift Packages' feature pointing to the HEAD of this feature branch to test :)

danielrbrowne:
I think once this is merged, a tagged release with fully working tvOS support and SPM integration on Xcode 12.0 and above 
should be possible!

PR #223, opened 2020-10-23
Add Aead support for ChaCha20Poly1305Ietf

JakobOffersen:
This PR extends `Aead` to also support `ChaCha20Poly1305Ietf`.

The extension + tests follows the code structure of the already implemented `XChaCha20Poly1305Ietf `


JakobOffersen:
@jedisct1 Please tell me if I should add/change/remove anything for this PR to be accepted 😊 
- Cheers

jedisct1:
The nonce size is a little bit short for nonces to be randomly chosen. So, we can't use the same interface.

Issue #224, opened 2020-11-04
Support for server side swift (Linux)


code28:
Using this new construct with libsodium shipped in the `Clibsodium.xcframeworks`, server side swift isn't supported anymore. When compiling, this error message occurs:
```
error: artifact 'Clibsodium' does not support the target platform and architecture ('Triple(tripleString: "x86_64-unknown-linux-gnu", arch: SPMBuildCore.Triple.Arch.x86_64, vendor: SPMBuildCore.Triple.Vendor.unknown, os: SPMBuildCore.Triple.OS.linux, abi: SPMBuildCore.Triple.ABI.unknown)')
```
We are using Sodium on our Swift server, which wasn't a problem before and everything worked well by installing libsodium globally.
It'd be great to have support for `x86_64-unknown-linux-gnu` architecture with this construct as well. Is this planned?

jedisct1:
Do you know how to do that?

code28:
I guess that the file [dist-build.sh](https://github.com/jedisct1/libsodium/blob/master/dist-build/apple-xcframework.sh) just needs to get another architecture - `x86_64-unknown-linux-gnu` - and the `xcframework` needs to be updated in this repo. I think, that should be enough. You are probably way more aware of what to write in that file, and I suppose it shouldn't be a lot of lines. Or do you see something, which could add complexity?

I'm happy to help further where I can - especially with trying, if it works.

jedisct1:
But how do you compile to Linux? Using Zig maybe?

code28:
Ah, I see what you mean. Actually didn't think about it before... I don't know a lot about Zig, so I don't know if it could help, but this could be done quite easily using Docker. This adds some overhead, because Docker must be installed, but otherwise, it should be quite easy:

Dockerfile:
```Dockerfile
FROM ubuntu:20.04

RUN export DEBIAN_FRONTEND=noninteractive DEBCONF_NONINTERACTIVE_SEEN=true \
    && apt-get -q update \
    && apt-get -q dist-upgrade -y\
    && apt-get install -y curl build-essential\
    && rm -rf /var/lib/apt/lists/*

WORKDIR /tmp
COPY . .
CMD \
  make distclean >/dev/null 2>&1 \
  ./configure && \
  make install || exit 1
```

In `apple-xcframework.sh`:
```shell
docker build -t libsodium-docker-builder .
docker run --rm -v `pwd`:/tmp libsodium-docker-builder
[...]

PR #225, opened 2020-11-23
Use xcframework only for supported OS

code28:
This is not quite a solution for the problem of #224, but it's a running workaround. Otherwise, swift-sodium can't be used on server side projects on Linux - this makes it possible again.

### DIFF
--- a/Clibsodium/module.modulemap
+++ b/Clibsodium/module.modulemap
@@ -1,0 +1,5 @@
+module Clibsodium [system] {
+  header "shim.h"
+  link "sodium"
+  export *
+}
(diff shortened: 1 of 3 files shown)


jedisct1:
That's a hack that may work around a current issue, but depending on another repository, and on different versions of the library doesn't seem like the best way to go.

code28:
Yeah, you're right, it's not the best way, but this works. And since binary targets aren't supported on linux, I don't see another way to get this done, except from this workaround.
Regarding the dependency, I think, this can be added inside this repository. I'll give it a try.

code28:
Now the change is even smaller and it doesn't depend on an external dependency.

PR #232, opened 2020-12-16
Some updates for README.md docs

johnalanwoods:
Adds some detail on cryptographic primitives, fixes some language and adds an example of variable length hash.

More work is still needed, however this is an improvement.


jedisct1:
Thank you!

`ed25519` is a signature system, not a curve. The underlying curve is `edwards25519`, but no worries, I can fix that.

Thanks again!

PR #234, opened 2020-12-21
Factor `Clibsodium` into its own sub-package.

### LINKED ISSUE #233: First build fails with "No such module 'Clibsodium'" when paired with CocoaLumberjack
When I create a framework build target which includes both `swift-sodium` and `CocoaLumberjack`, the first build always fails on `Aead.swift` with the error "No such module 'Clibsodium'".  The second build succeeds.  This repeats any time I clean the workspace.   The first build in a clean environment always fails.

The trouble is that this prevents me from archiving the framework, since the archive operation always performs a clean at the beginning.

I wonder if there might be a problem with parallel builds here in which the `Sodium` target is trying to reference `Clibsodium` before it is ready, and building the `CocoaLumberjack` module triggers this timing condition.

I've attached a project demonstrating this issue.  This project contains a framework, not an app, and I've disabled code signing.  So you should be able to download and reproduce the issue without any modifications to the project.

[SodiumBuildTest.zip](https://github.com/jedisct1/swift-sodium/files/5725900/SodiumBuildTest.zip)

This might be related to #226

jnross:
This change appears to fix #233, which may occur due to Xcode incorrect parallelizing the build of target dependencies in a single package.  Putting Clibsodium in its own package appears make this dependency more clear.

I would mark this PR as a draft if I could.  In particular I have not tested whether my change to the podspec works correctly.  That part of the diff is just a guess.


Sodium.podspec
```diff
@@ -32,5 +32,5 @@ Pod::Spec.new do |s|
 
   s.requires_arc = true
 
-  s.vendored_frameworks = "Clibsodium.xcframework"
+  s.vendored_frameworks = "Clibsodium/Clibsodium.xcframework"
```
jnross, line 35: I'm not sure if this change to the podspec works correctly with the directory structure changes.

jnross:
This approaches works fine locally, but throws some weird errors about versioning when I try to add the dependency pointed to the remote git URL.

jnross:
Closing.  This is not what `Package(path:)` is for.  Right now the only approach that I know works is to make Clibsodium a separate package in a separate repo.

PR #238, opened 2021-01-12
Circumvent swift bug SR-14038

code28:
Unfortunately, a swift bug makes it impossible to use swift-sodium on linux. The fix is really quick and easy.

The bug is the following: https://bugs.swift.org/browse/SR-14038


jedisct1:
Thank you!

PR #244, opened 2021-03-07
Make initializer throwable, update .gitignore

own2pwn:
IMO it's better to expect initializer to throw rather than throw silently.
Updated gitignore to match auto-generated one by Xcode (unsure, if `build` should be `/.build` or these are distinct here)

### DIFF
--- a/Sodium/Sodium.swift
+++ b/Sodium/Sodium.swift
@@ -30,7 +30,7 @@ public struct Sodium {
 }
 
 extension Sodium {
-    private static let initCode: Int32 = {
+    private static let initCode: CInt = {
         return sodium_init()
     }()
 }
--- a/Sodium/Sodium.swift
+++ b/Sodium/Sodium.swift
@@ -1,7 +1,11 @@
-import Foundation
 import Clibsodium
+import Foundation
 
 public struct Sodium {
+    public enum InitError: Error {
+        case code(Int32)
+    }
+
     public let box = Box()
     public let secretBox = SecretBox()
     public let genericHash = GenericHash()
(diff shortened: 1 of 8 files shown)


jedisct1:
It's never going to throw on Apple devices.

own2pwn:
Closing? @jedisct1

Sodium/Sodium.swift
```diff
@@ -17,15 +21,16 @@ public struct Sodium {
...
     }
 }
 
 extension Sodium {
-    private static let once: Void = {
-        guard sodium_init() >= 0 else {
-            fatalError("Failed to initialize libsodium")
-        }
+    private static let initCode: Int32 = {
+        return sodium_init()
```
jedisct1, line 34: `sodium_init()` returns a `Cint`, not an `Int32`. But the return code can only be `-1` on error.

Tests/SodiumTests/ReadmeTests.swift
```diff
@@ -28,7 +28,7 @@ class ReadmeTests : XCTestCase {
     ]
 
     func testAuthenticatedEncryption() {
-        let sodium = Sodium()
+        let sodium = try! Sodium()
```
jedisct1, line 31: That change is probably breaking all the example applications as well.

own2pwn: My bad. Fixed

Sodium/Sodium.swift
own2pwn: Replaced with `CInt`. It is `Int32` on my mac (unsure about Linux, but probably that's independent)

own2pwn: I kept the previous check for `code >= 0`. Should we change it to `if code == -1 { throw InitError.code(-1) }` ?

jedisct1: No, but what I meant is that the return code is not something worth propagating in `InitError`. The code is always going to be `-1`.

jedisct1:
I'm not sure about this...

It breaks all applications. And if initialization fails, there's not much an application can do besides crash or quit.

own2pwn:
That's quite a point, @jedisct1. I don't mind if we just close this

PR #249, opened 2021-09-29
Full libsodium build

### LINKED ISSUE #248: Clibsodium without `SODIUM_LIBRARY_MINIMAL`?
Would there be any issue, in principle, with adjusting the precompiled libsodium to be a non-minimal build? At the moment I believe that `SODIUM_LIBRARY_MINIMAL 1` is set, and so some parts of Clibsodium are not available.

i.e. all of these are not currently available when importing `Clibsodium`

https://github.com/jedisct1/swift-sodium/blob/1d51e8eb8a99638166aea863e7a71d568b5a06e6/Sodium/libsodium/sodium_lib.h#L56-L67

aidantwoods:
I re-ran dist-build/apple-xcframework.sh from libsodium with `LIBSODIUM_FULL_BUILD=1`, and manually corrected the file `Sodium/libsodium/version.h` to match. For the sake of simplicity about what actually changed, I also re-ordered the `AvailableLibraries` array in `Clibsodium.xcframework/Info.plist` from what was output by this script so as to match the current file ordering (and therefore this file is now unchanged).

This was run using Xcode 12.5.1 on revision 7d71804ee05f71b65c94282f706c5aa82b279e8a of libsodium (as currently documented in the readme).

This exposes some currently missing functionality from Clibsodium https://github.com/jedisct1/swift-sodium/blob/1d51e8eb8a99638166aea863e7a71d568b5a06e6/Sodium/libsodium/sodium_lib.h#L56-L67

Even though this doesn't change the swift library, I think it makes sense for this package to make all of libsodium available in-case any of the low level functionality is needed?

I've opened this PR for convenience, but I won't be offended if you prefer to build the modified binaries yourself  :)

Closes #248


johnalanwoods:
nice, this would be useful.

KINGH242:
I need to use `crypto_scalarmult_ed25519` in a project. Is there any chance this will get merged or are there any other options?

PR #255, opened 2022-05-27
added: SodiumRandomNumberGenerator

c128128:
A more swifty way of generating random numbers.

```swift
var rng = SodiumRandomNumberGenerator()
let randomUInt32 = UInt32.random(in: 0...10, using: &rng)
let randomUInt64 = UInt64.random(in: 0...10, using: &rng)
let randomInt = Int.random(in: 0...10, using: &rng)
let randomDouble = Double.random(in: 0...1, using: &rng)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,10 +248,10 @@ let seed = "0123456789abcdef0123456789abcdef".bytes
 let stream = sodium.randomBytes.deterministic(length: 1000, seed: seed)!
 ```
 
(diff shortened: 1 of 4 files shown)


jedisct1:
Instead of introducing another structure in the same namespace as `RandomBytes`, could it be returned by something like `RandomBytes.generator`?

c128128:
How about to conform `RandomBytes` to `RandomNumberGenerator` ?

jedisct1:
Thank you so much for this!

PR #256, opened 2022-06-05
add Aead.XChaCha20Poly1305Ietf encrypt function which takes a Nonce as a parameter

clipsyncdev:
The goal of this change is to add a version of the swift `sodium.aead.xchacha20poly1305ietf.encrypt` that is a little more similar to the JS `sodium.crypto_aead_xchacha20poly1305_ietf_encrypt` which takes a nonce as a passed parameter instead of generating it in the function and then returning it.


aidantwoods:
Related discussion: https://github.com/jedisct1/swift-sodium/pull/173

clipsyncdev:
> Related discussion: #173

This is highly unlikely to get merged based on that discussion.

PR #266, opened 2024-04-24
Add test vectors from RFC8032 for ed25519 signatures

alinradut:
Added test vectors for ed25519 from RFC8032: https://datatracker.ietf.org/doc/html/rfc8032#section-7.1


blochberger (approved):
The actual test code is repetitive, so it could be done with a data-driven approach, but LGTM otherwise. The test vectors match with the RFC.

jedisct1:
Thank you!

PR #268, opened 2024-08-04
Gardening for SwiftPM

sinoru:
This PR includes improvements for SwiftPM contains:
- Better platform determination which should use XCFramework by build target than host
- Make Xcodeproj generating Info.plist instead of repo storing
- Remove `Sodium.h` which contains `SodiumVersionNumber`, `SodiumVersionString[]` symbols only generated by Xcode builds
- Remove `LinuxMain.swift` which is [deprecated](https://github.com/swiftlang/swift-package-manager/blob/1284494342d1c8e413bd7744bdc432a8a75bf165/Documentation/ReleaseNotes/5.4.md?plain=1#L16)

I've been test in Xcode 15.4, Swift 5.10 on macOS, swift:5.10-bookworm docker image (with libsodium 1.0.20).


diederich:
+1 on this. I like moving to a more standard package layout and using the new machinery for platform dependencies! 🚀

PR #272, opened 2025-08-05
Add support for SHA256

tristanlabelle:
Hello! We have a fork that exposes SHA256 hashing following conventions from this repo. We'd be happy to contribute it to the main project so we can use it directly.

PR #273, opened 2025-08-11
Exposes SHA-256 and SHA-512

tristanlabelle:
Exposes the SHA-256 and SHA-512 hashing algorithms, with accompanying tests and documentation following codebase standards.


Tests/SodiumTests/SodiumTests.swift
```diff
@@ -270,6 +272,18 @@ class SodiumTests: XCTestCase {
...
+        let message = "My Test Message".bytes
+        let h = sodium.utils.bin2hex(sodium.sha256Hash.hash(message: message)!)!
+        XCTAssertEqual(h, "06b87ce47e9cb96156a4f55e405c2587fff2be2ed11adb9f465dc964ef70938e")
+    }
+    
+    func testSHA512Hash() {
+        let message = "My Test Message".bytes
+        let h = sodium.utils.bin2hex(sodium.sha512Hash.hash(message: message)!)!
+        XCTAssertEqual(h, "4173c1218c42b37943f39a7c5a87b782ae945b8609b8dae66a0552c56abd931ba0561a6df052008c1994a505024c31f183a5c34c0dda1e0ce82e9b9e7ea08459")
+    }
```
blochberger, line 286: I would align the tests with libsodium for consistency and to ensure that the values are identical across language boundaries, see [`libsodium/test/default/hash.c`](https://github.com/jedisct1/libsodium/blob/59b8ba3573b6cde4c2dfca96326175c30b07961b/test/default/hash.c).

blochberger (commented):
Change looks good to me otherwise.

blochberger:
Could have been pushed to #272, I suppose.

tristanlabelle:
@blochberger , I had to switch the source branch so I couldn't reuse the PR. I meant to close the original one but was at the airport and my boarding group got called at the exact moment I published this lol.

Tests/SodiumTests/SodiumTests.swift
tristanlabelle: I'm not finding any test calling `crypto_hash_sha256` and asserting on the resulting hash in that file or in libsodium in general, only resulting bytes getting printed and assertions about hash digest sizes. Am I missing something? Or would you like me to grab the printed output from hashing `static unsigned char x[] = "testing\n";` and assert against it here? All the other tests in `SodiumTests.swift` use `let message = "My Test Message".bytes` so this seemed like the most consistent way to write the new tests.

tristanlabelle: I added the test case from libsodium to the sha256 test. There was none that I could find for sha512.

blochberger: `crypo_hash()` is  called in the linked test, which defaults to `crypto_hash_sha512()`, i.e., the first two lines of the `hash.exp` file contain the expected SHA-512 hashes for the values of `x` and `x2` respectively.

tristanlabelle: Ah I see. Added in latest.

tristanlabelle:
@blochberger Please let me know if we're all good, or if I need to ping the maintainer of the repo to get this merged!

Sodium/Sodium.swift
```diff
@@ -9,6 +9,8 @@ public struct Sodium {
     public let randomBytes = RandomBytes()
     public let shortHash = ShortHash()
     public let sign = Sign()
+    public let sha256Hash = SHA256Hash()
+    public let sha512Hash = SHA512Hash()
```
blochberger, line 13: I'm wondering about the API. Seems like this deviates from the general pattern of swift-sodium compared to libsodium. How about adding a `cryptoHash` layer that encapsulates the SHA-256/512 implementations. Feels a bit odd to have very specific algorithms exposed at the top, given the main idea behind libsodium to offer high-level interfaces.

tristanlabelle: Ping @blochberger . This feels close to landing and I'd like your help to get it there.

tristanlabelle: Re-ping @blochberger

jedisct1: I agree that having these functions at the top-level feels very odd. They are low-level primitives, and not high-level functions.

Ideally, there should be a new namespace for primitives, with another namespace for hash functions inside.

Also, SHA512-256 is missing even though this is the safest SHA2-based option.

jedisct1: > > I can't find this functionality in the C code. There's `crypto_auth_hmacsha512256` but that's hmac.
> Apparently a SHA512-256 is not just a truncated SHA512 so I won't risk an implementation if it's not exposed.

My bad, I thought `crypto_hash()` was truncated SHA512 like `crypto_auth()` so it's not vulnerable to length extensions attacks, but it's indeed just SHA512.

tristanlabelle: All right. Could you also confirm if the commit I linked is the changes you need to merge this PR, and otherwise provide more info or push a commit doing it?

jedisct1: I was rather thinking of something like `primitives.hash.sha256` .

